PR #3, opened 2019-06-25
Initial add of jenkinsfile

limed:


Jenkinsfile
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  stage ('Buld Image'){
+    python_app = docker.build('mdnwebdocs/developer-portal-app:${tag}', '-f ./dockerfiles/Dockerfile.app .')
+    node_app = docker.build('mdnwebdocs/developer-portal-static:${tag}', '-f ./dockerfiles/Dockerfile.static .')
+  }
+
+  stage('Push Image') {
+    printlin("Pushing images")
+    python_app.push()
```
escattone: Oh, it looks like you're not using the `tag` now, so the `.push()` must do that be default? So maybe we no longer need the `get_commit_tag` function?

Jenkinsfile
```diff
@@ -0,0 +1,31 @@
...
+    checkout scm
+    tag = get_commit_tag()
+  }
+
+  stage ('Buld Image'){
+    python_app = docker.build('mdnwebdocs/developer-portal-app:${tag}', '-f ./dockerfiles/Dockerfile.app .')
+    node_app = docker.build('mdnwebdocs/developer-portal-static:${tag}', '-f ./dockerfiles/Dockerfile.static .')
+  }
+
+  stage('Push Image') {
```
escattone: Before we push the images, I think we should ensure that these tests pass:
```
docker-compose exec app python manage.py test
docker-compose exec static npm run test
```
but perhaps you were planning to do that later in a separate PR.

Jenkinsfile
```diff
@@ -0,0 +1,31 @@
+
+def get_commit_tag() {
+  return sh(returnStdout: true, script: "git log -n 1 --pretty=format:'%h'").trim()
```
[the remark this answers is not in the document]
limed: That works for me, that was copy and pasted somewhere I like your method better. I think kuma uses the first 7 characters so perhaps we should use that as well

Jenkinsfile
limed: We can be explicit here but when you don't specify anything it pushes the image that we built, in this case its `docker.build('mdnwebdocs/developer-portal-app:${tag}', '-f ./dockerfiles/Dockerfile.app .')`

Which means it pushes the tagged name and the pushes the `latest` tag

Jenkinsfile
limed: I was planning a separate PR for this. The docker-compose file will need to be modified to take the new images so lets make sure it gets built and pushed first

PR #42, opened 2019-08-01
External article authors

GaryHomewood:


dcgauld:
Yeah, they are used but there's such little content in there at the moment I don't think it's a big issue. I'm assuming a migration for this would be tricky?

PR #47, opened 2019-08-01
Tweaking build

limed:


limed:
I realize that the tests has failed here but its a known thing that the test is failing. Getting this in here so that we can at least have some travis tests setup

PR #49, opened 2019-08-02
Add description to directory pages

GaryHomewood:


developerportal/apps/articles/templates/articles.html
```diff
@@ -8,6 +8,7 @@
       {% if page.articles %}
         <section>
           <h2>Articles</h2>
+          <p class="page-description">{{ page.description }}</p>
```
dcgauld: We should probably have conditional here to only show the `<p>` tag if there's a description, as below.

GaryHomewood: Good point, done.

PR #53, opened 2019-08-02
Reverting c594f31c999088f28a2caf66730915f852428cab

limed:


limed:
Revert of commit c594f31c999088f28a2caf66730915f852428cab

limed:
r=limed

stevejalim:
Heya @limed! Could you clue me in about why this was reverted, please? Looks like it makes sense to have that SSL redirect in place -- is the revert because there's a different/better point in the stack where this is applied instead?

Cheers,
Steve

limed:
> Heya @limed! Could you clue me in about why this was reverted, please? Looks like it makes sense to have that SSL redirect in place -- is the revert because there's a different/better point in the stack where this is applied instead?
> 
> Cheers,
> Steve

Hey @stevejalim I believe when I initially deployed that it caused a redirect loop which is why I took that out. However I would love to actually have auto redirect to SSL enabled without the redirect loop. If you know of a way I am happy to make that happen

PR #54, opened 2019-08-02
Add video page type and directory

liamnewmarch:


developerportal/apps/people/templates/people.html
```diff
@@ -20,5 +20,8 @@ <h2>People</h2>
         {% include "organisms/newsletter-signup.html" %}
       </section>
     </div>
+    <section id="newsletter" class="section-full section-background">
```
dcgauld: Can you revert this change? Looks like a bad merge.

liamnewmarch: Removed

PR #58, opened 2019-08-05
Nav rename to communities, add MDN link

GaryHomewood:


developerportal/templates/header.html
```diff
@@ -20,9 +20,13 @@
                   {% for page in home_page.get_children.in_menu %}
                     {% with child_pages=page.get_children.in_menu %}
                       <li class="mzp-c-menu-category{% if child_pages %} mzp-has-drop-down mzp-js-expandable{% endif %}">
+                        {% if child_pages %}
+                        <span class="mzp-c-menu-title" >{{ page.title }}</span>
+                        {% else %}
                         <a class="mzp-c-menu-title" href="{{ base_url }}{% pageurl page %}" {% if child_pages %}aria-haspopup="true" aria-controls="mzp-c-menu-panel-{{ page.slug }}"{% endif %}>
```
liamnewmarch: I think we need these attributes on the span since they’re accessibility hints for the sub-nav.

GaryHomewood: Good catch, added to span.

PR #59, opened 2019-08-05
Add internal video template and styles

dcgauld:


developerportal/templates/molecules/article-details.html
```diff
@@ -1,20 +1,32 @@
...
-        height="{{ author_pic.height }}"
-        alt="{{ author_pic.alt }}"/>
+        src="{{ first_author_image.url }}"
+        width="{{ first_author_image.width }}"
+        height="{{ first_author_image.height }}"
+        alt="{{ first_author_image.alt }}"
+      />
     {% else %}
-      <img src="https://via.placeholder.com/60"/>
+      <img src="https://via.placeholder.com/60" />
```
GaryHomewood: Should we have a placeholder from assets instead?

PR #67, opened 2019-08-07
Add external support to event cards

dcgauld:


src/css/molecules/card-event.scss
```diff
@@ -104,20 +109,21 @@
     border-bottom-color: transparent;
     border-left-color: transparent;
     border-top-color: transparent;bottom: -14px;
```
GaryHomewood: Need a line break here?

developerportal/apps/externalcontent/models.py
```diff
@@ -165,7 +166,25 @@ def event(self):
...
+            if self.end_date.strftime("%m") == start_month:
+                event_dates += self.end_date.strftime("%-d")
+            else:
+                event_dates += self.end_date.strftime("%b %-d")
+        return event_dates
+
+    @property
+    def event_dates_full(self):
+        """Return a formatted string of the event start and end dates, including the year"""
+        return self.event_dates + self.start_date.strftime(", %Y")
```
GaryHomewood: This duplication between on site and external models could be cleaner.

dcgauld: Agreed. I've raised a ticket to cover the creation of abstract classes above these, to add common helpers and fields etc.

PR #68, opened 2019-08-07
Show videos on topic page

GaryHomewood:


developerportal/apps/videos/models.py
```diff
@@ -92,7 +92,7 @@ class Video(Page):
     template = 'video.html'
 
     # Content fields
-    description = CharField(default='', blank=True, max_length=250)
+    description = TextField(default='', blank=True, max_length=250)
```
dcgauld: Out of interest, what's this change for? And if needed, does there need to be a related migration?

GaryHomewood: Made it a textfield to allow multi-line entry in the admin. Have now included the necessary migrations.

PR #78, opened 2019-08-08
404 and 500 error pages

GaryHomewood:


src/css/templates/http-error.scss
```diff
@@ -0,0 +1,67 @@
+.full-height {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  min-height: 100vh;
```
dcgauld: By using `100vh`, do we need to take into account the navigation's height too? Otherwise there may be scrollbars?

PR #84, opened 2019-08-12
remove restart:always in docker-compose

### LINKED ISSUE #83: `restart: always` in docker-compose is "annoying"
I work on several projects (but primarily kuma) that uses `docker-compose`. 
From the first time I cloned `developer-portal` for the first time and ran `docker-compose up` now, on my docker-for-mac, it thinks I want to start the `developer-portal` project every time I restart docker (e.g. rebooting my computer). 

[It happens because of this line](https://github.com/mdn/developer-portal/blob/c9c6de22ee8a9a2fd69740ffb4475bb368742f2a/docker-compose.yml#L9)

If the people who work on `developer-portal` every workday of the week, want this feature I think it would be best to use [`docker-compose.override.yml`](https://docs.docker.com/compose/reference/overview/).

peterbe:
Fixes #83

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,6 @@ services:
     expose:
       - 5432
     image: postgres:11.3-alpine
-    restart: always
     tty: true
     volumes:
       - database:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


limed:
r+ on my end

dcgauld:
Thanks @peterbe 👍

PR #85, opened 2019-08-12
Add optional hero image to content page template

GaryHomewood:


developerportal/apps/content/templates/content.html
```diff
@@ -6,18 +6,32 @@
 {% block body_class %}content{% endblock %}
 
 {% block content %}
-  <main>
+  <main class="{% if page.hero_image %}content-has-hero{% endif %}">
```
dcgauld: We should be able to use the `has-header` class for this.

developerportal/apps/content/templates/content.html
```diff
@@ -6,18 +6,32 @@
...
-        {% for block in page.body %}
-          {% if block.block_type == 'image' %}
-            {% image block.value width-672 as rendition %}
-            {% include "atoms/image.html" with image=block.value rendition=rendition %}
-          {% else %}
-            {{ block }}
-          {% endif %}
-        {% endfor %}
+        {% if page.hero_image %}
+          {% image page.hero_image width-1024 as hero %}
```
dcgauld: Do we need to support retina images here? If so, maybe this is a good opportunity to use `srcset` since serving a 2048px wide image to a mobile device wouldn't be cool.

developerportal/apps/content/templates/content.html
```diff
@@ -6,18 +6,32 @@
...
+          </div>
+        {% else %}
+          <h1>{{ page.title }}</h1>
+        {% endif %}
+      </section>
+      <section class="article-content">
+        <article class="article-body">
+          {% for block in page.body %}
+            {% if block.block_type == 'image' %}
+              {% image block.value width-672 as rendition %}
```
dcgauld: Sorry, just noticed perhaps this image size should be updated? I think maybe this is why they're having image issues with this particular page type.

developerportal/apps/content/templates/content.html
GaryHomewood: Ok done.

developerportal/apps/content/templates/content.html
GaryHomewood: To be addressed as separate responsive image ticket

developerportal/apps/content/templates/content.html
GaryHomewood: Content is now centred so this rendition size is now correct.

PR #89, opened 2019-08-13
Add button block and content page fixes

dcgauld:


developerportal/apps/common/blocks.py
```diff
@@ -14,6 +24,7 @@ class CodeSnippetBlock(blocks.StructBlock):
     code = blocks.TextBlock()
 
     class Meta:
+        icon = 'code'
```
GaryHomewood: This is nice. Makes it author friendly.

PR #92, opened 2019-08-13
Add help text to models

liamnewmarch:


developerportal/apps/common/constants.py
```diff
@@ -36,6 +36,12 @@
     ('purple-20', '#f68fff'),
 )
 
+COUNT_CHOICES = (
+    (3, '3'),
+    (6, '6'),
+    (9, '9'),
+)
+
```
GaryHomewood: Could be a more specific name? LATEST_ARTICLES_COUNT?

liamnewmarch: Didn’t want to limit this to ‘latest’ or ‘article’ in case it’s used elsewhere, discussed IRL and settled on the marginally clearer RESOURCE_COUNT_CHOICES. Naming things is hard!

PR #94, opened 2019-08-14
Add darker hero to homepage

dcgauld:


src/css/organisms/homepage-header.scss
```diff
@@ -1,24 +1,29 @@
 .homepage-header {
-  background: url('/static/img/homepage-header-pattern.svg') $color-gray-20;
+  background: url('/static/img/homepage-header-pattern.svg') $color-black;
   background-position: calc(50% + 400px);
```
liamnewmarch: Just out of curiosity, why not roll these lines into the `background` above? Compatibility for `calc()`?

dcgauld: Good shout! Updated 👌

PR #95, opened 2019-08-14
Use accent color derived from primary topic with fallback

liamnewmarch:


developerportal/templates/utils/accent-color.html
```diff
@@ -1,4 +1,5 @@
 {% load app_filters %}
+{% firstof color '#88ffd1' as color %}{# Set fallback color #}
```
dcgauld: I think this default might need to be set within the event template itself, since the default color changes based on the section (I think `#88ffd1` is only the default for the events section for instance).

The main default for the whole site is yellow, which is defined within the main CSS files.

liamnewmarch: Nice, I didn’t catch that. Updated.

PR #102, opened 2019-08-14
WIP: IE fixes

GaryHomewood:
- Layout and JS on directory pages
- Layout for card container

### DIFF
--- a/src/css/molecules/card.scss
+++ b/src/css/molecules/card.scss
@@ -20,9 +20,9 @@
     margin-right: $grid-gap;
   }
 
-  .card-featured {
-    max-width: 494px;
-    margin-right: $grid-gap;
+  .card-featured,
+  .card.external-promo {
+    max-width: 480px;
   }
 
   .card-link {
@@ -54,6 +54,10 @@
 
     .card {
       max-width: 320px;
+    }
+
+    .card.external-promo {
+      max-width: 100%;
     }
 
     .card-featured {
--- a/src/css/molecules/card.scss
+++ b/src/css/molecules/card.scss
@@ -45,6 +45,7 @@
     display: grid;
     grid-gap: $grid-gap-mobile;
     grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
+    max-width: unset;
 
     @media #{$mq-md} {
       grid-column-gap: $grid-gap;
--- a/src/css/molecules/card.scss
+++ b/src/css/molecules/card.scss
@@ -1,16 +1,83 @@
+.section-header,
 .card-container {
-  display: grid;
-  grid-gap: $grid-gap-mobile;
-  grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
+  margin-left: auto;
+  margin-right: auto;
+  padding: 0;
+  position: relative;
 
   @media #{$mq-md} {
-    grid-column-gap: $grid-gap;
-    grid-row-gap: 24px;
+    max-width: 1024px;
   }
 }
 
-.card-container-2-wide {
-  grid-template-columns: repeat(auto-fill, minmax(340px, 1fr));
+.card-container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+
+  .card {
+    max-width: 300px;
+    margin-right: $grid-gap;
+  }
+
+  .card-featured {
+    max-width: 494px;
+    margin-right: $grid-gap;
+  }
+
+  .card-link {
+    flex-grow: 1;
+    margin-bottom: 24px;
+  }
+}
+
+.filter-list-sidebar {
+  margin-right: $grid-gap;
+}
+
+.item-topic-link {
+  max-width: 320px;
+  margin-bottom: 24px;
+}
+
+@supports (grid-gap: $grid-gap-mobile) {
+  .card-container {
+    display: grid;
+    grid-gap: $grid-gap-mobile;
+    grid-template-columns: repeat(auto-fill, minmax(250px, 1fr));
+
+    @media #{$mq-md} {
+      grid-column-gap: $grid-gap;
+      grid-row-gap: 24px;
+    }
+
+    .card {
+      max-width: 320px;
+    }
+
+    .card-featured {
+      max-width: unset;
+      margin-right: unset;
+    }
+
+    .card-link {
+      flex-grow: 1;
+      margin-bottom: unset;
+    }
+  }
+
+  .card-container-2-wide {
+    grid-template-columns: repeat(auto-fill, minmax(340px, 1fr));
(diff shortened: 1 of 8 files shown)


src/css/organisms/filter-list.scss
```diff
@@ -1,5 +1,7 @@
 .filter-list-content {
+  display: -ms-grid;
```
liamnewmarch: Is this not added by autoprefixer?

src/js/organisms/filter-form.js
```diff
@@ -51,7 +51,7 @@ module.exports = class FilterForm {
     });
 
     this.nextPageButton.addEventListener('click', e => this.nextPage(e));
-    this.form.addEventListener('input', () => this.onFormInput());
+    this.form.addEventListener('change', () => this.onFormInput());
```
liamnewmarch: I didn’t know [older browsers and pre-Blink Edge don’t trigger `input` on checkboxes][1], nice catch!

[1]: https://caniuse.com/#feat=input-event

src/js/organisms/filter-form.js
```diff
@@ -89,7 +89,15 @@ module.exports = class FilterForm {
     });
 
     if (matchedCheckboxes.length) {
-      this.form.dispatchEvent(new Event('input'));
+      let event;
+      if (typeof(Event) === 'function') {
```
liamnewmarch: `typeof` is a keyword so you don’t need the parentheses around `Event`. `typeof Event === 'function'` works the same.

Could this be shimmed like the other methods in pollyfills.js? Something like:

```
if (!window.Event) {
  window.Event = function(typeArg) {
    const event = document.createEvent('Event');
    event.initEvent(typeArg, true, true);
    return event;
  };
}
```

I think that should work for this case, but [proper pollyfills are available][1] too.

[1]: https://github.com/webcomponents/polyfills/tree/master/packages/webcomponentsjs

src/js/polyfills.js
```diff
@@ -0,0 +1,102 @@
+// Production steps of ECMA-262, Edition 6, 22.1.2.1
```
liamnewmarch: Do we need to give credit for these polyfills? If they’re available on npm it might be worth adding them to our build pipeline.

src/js/utils.js
```diff
@@ -17,12 +17,28 @@ exports.parseQueryParams = () => {
...
 exports.parseForm = form => {
   const formData = new FormData(form);
-  return Array.from(formData.entries()).reduce((acc, [key, value]) => {
-    if (!acc[key]) {
-      acc[key] = [];
+  if (!formData.entries) {
+    // for IE
+    let filter = {};
+    const elements = document.getElementsByTagName("input");
+    for (let i = 0; i < elements.length; i++) {
```
liamnewmarch: Minor, but since there’s a polyfill for `Array.from` now, you could probably do

```
Array.from(elements).forEach((element) => {
  // Do things with element
});
```

src/js/utils.js
```diff
@@ -17,12 +17,28 @@ exports.parseQueryParams = () => {
 /** Creates an object based on a form's current input values. */
 exports.parseForm = form => {
   const formData = new FormData(form);
-  return Array.from(formData.entries()).reduce((acc, [key, value]) => {
-    if (!acc[key]) {
-      acc[key] = [];
+  if (!formData.entries) {
+    // for IE
+    let filter = {};
+    const elements = document.getElementsByTagName("input");
```
liamnewmarch: You could offload some later work to the browser by using a selector here

```
const elements = document.querySelectorAll('input[type="checkbox"]:checked');
```

src/js/utils.js
```diff
@@ -17,12 +17,28 @@ exports.parseQueryParams = () => {
 /** Creates an object based on a form's current input values. */
 exports.parseForm = form => {
   const formData = new FormData(form);
-  return Array.from(formData.entries()).reduce((acc, [key, value]) => {
-    if (!acc[key]) {
-      acc[key] = [];
+  if (!formData.entries) {
```
liamnewmarch: Minor, but you could check `if (FormData && FormData.prototype.entries)` to avoid constructing FormData which might not exist or exists but then the instance isn’t used.

src/css/organisms/filter-list.scss
liamnewmarch: Ah! It looks like autoprefixer [handles grid support slightly differently][1]. We might also need to adjust the browserlist in package.json to include IE.

[1]: https://github.com/postcss/autoprefixer#grid-autoplacement-support-in-ie

GaryHomewood: The autoprefixer didn't work as intended for this scenario, so I set the rules explicitly to get the layout required.

src/js/organisms/filter-form.js
GaryHomewood: Me neither. Took a while to find!

src/js/utils.js
GaryHomewood: That's a good idea, done.

src/js/utils.js
GaryHomewood: Yes that makes it cleaner, done.

src/js/utils.js
GaryHomewood: Yes, that makes it cleaner, done.

src/js/organisms/filter-form.js
GaryHomewood: Added a shim as suggested.

src/js/polyfills.js
GaryHomewood: Added source links where appropriate.

src/js/utils.js
```diff
@@ -16,13 +16,24 @@ exports.parseQueryParams = () => {
 
 /** Creates an object based on a form's current input values. */
 exports.parseForm = form => {
-  const formData = new FormData(form);
```
dcgauld: I'm just thinking, if we have to have two implementations then maybe we should _just_ use the non-FormData one?

src/css/molecules/card.scss
```diff
@@ -1,18 +1,85 @@
...
+.filter-list-sidebar {
+  margin-right: $grid-gap;
+}
+
+.item-topic-link {
+  max-width: 320px;
+  margin-bottom: 24px;
 }
 
+@supports (grid-gap: $grid-gap-mobile) {
```
dcgauld: Can we just check for `display: grid` here?

GaryHomewood: Sorry yes, I thought IE supported grid, buts its -ms prefixed.

src/js/utils.js
GaryHomewood: Agree, that makes it more maintainable.

PR #122, opened 2019-08-19
Add toggle component and update speaker cards

dcgauld:


developerportal/templatetags/app_tags.py
```diff
@@ -12,7 +12,7 @@
 
 
 @register.simple_tag
-def make_list(*items):
+def make_list_from_args(*items):
```
liamnewmarch: Does this tag name still need to change if you’re using a filter?

PR #130, opened 2019-08-20
Ensure StreamBlocks display on content pages

dcgauld:


dcgauld:
Merging without a review as this is a fairly urgent bug fix.

PR #171, opened 2019-09-20
Fixing a typo in makefile

limed:
This changeset addresses a typo in the Makefile which cause a persistent
volume and persisten volume claim to not be able to delete

## Key changes:

- Added "pv" and "pvc" keyword in makefile to actually allow deletion to
happen


limed:
r=limed

PR #196, opened 2019-09-20
Chore/reconfigure renovate

stevejalim:
Tuning renovate - leave docker alone, automatically rebase, create a master PR


stevejalim:
It sounds like we won't need to merge this PR (https://docs.renovatebot.com/reconfigure-renovate/) but when Renovate has noticed it, we should to keep the config current, I think

PR #204, opened 2019-09-23
133: Address problem of static/S3-site URLs linking back to Wagtail live views

### LINKED ISSUE #133: Configure build to target the stage developer-portal published domain
The developer-portal `build` command needs to be configured for stage such that the `BASE_URL` setting is https://developer-portal-published.stage.mdn.mozit.cloud (or something to that effect). This issue arose out of the work done on https://github.com/mdn/developer-portal/issues/99.

stevejalim:
This changeset (when combined with a correct configuration for Site.hostname in Wagtail) fixes the issue where the static/baked site was linking back to live wagtail CMS instead of to itself.

The previous swing at this (#195) ~was too complex, and~ had a shortcoming, so I closed it.

Closes #133

--

**Changes:**

* Removed the use in templates of a `base_url` value that points to the Wagtail CMS. This now means that the URLs in any page (live-on-Wagtail or "baked") are relative rather than absolute, which means they are always appropriate for the site being browsed, whether it's the live Wagtail-hosted one or the S3-hosted baked site.

* In addition, for the moments where the hostname _is_ required to build an absolute URL, the main `Site.hostname` value should be set to the URL of the site where the baked site will be deployed, rather than the CMS URL where it's being made and/or live-viewed. (Wagtail can still serve live pages on that CMS URL, even when the hostname is set to the S3 site's domain).

* Related to the above, there is one area (Open Graph `<meta>` tags) where absolute URLs are definitely needed. For this, I've just added a simple template tag which uses the `HttpRequest` to get the scheme and domain, which is set the first`Site`'s `hostname` when the site is being baked out for copying to S3.

* Add support for an explicit EXPORTED_SITE_URL in settings

    Please read commit b314371 for details

* Update k8s staging config with EXPORTED_SITE_URL

    IMPORTANT: given that `stage.sh` mentions `dev` in the value used for `BASE_URL` while `stage-oregon.sh` mentions `stage` for that variable, I've basically made up a new FQDN for the `stage.sh` dev config, even though it doesn't exist yet (which is "OK", because the dev deployment appears not to be up anyway). Having both of the configurations point to the same published/static staging FQDN didn't seem sensible, because of the confusion it would cause. I'm totally open to being told this isn't legit - please do shout if so 😄 

**TLDR:** 
* I've stopped us using absolute URLs based on the `Site` and made them all relative, apart from OG tags, which do need to be absolute.
* Those absolute URLs are made via a templatetag, from the `HTTPRequest`. That `HTTPRequest` is only correctly configured if `Site.hostname` is set to the value of the S3 bucket's hostname, _not_ Wagtail's operating hostname. 
* I've added an extra configuration variable to allow us to export a static site that's not on the same hostname as the Wagtail CMS (ie, `EXPORTED_SITE_URL` must align with whatever `Site.hostname` is now set to)

**Post-deploy actions**
- [ ] Update the `hostname` of the site in the deployment to refer to the S3 bucket's hostname not the one Wagtail is on.

## How to test

- check out local build
- update the hostname of the Site in Wagtail admin to be something different from the default `localhost` and save
- confirm the CMS site still responds (bounce docker then reload some CMS views, view live pages, etc) -- the Site change won't have broken that.
- edit a page where there's a Card to include an image (eg a `Person` page - the card image will appear in Open Graph `<meta>` links.)
- publish the page
- view the live page for that page with the Card associated and view-source to confirm the `og` tags have absolute URIs in them, while all `a` tags have a relative `href`. eg, here's the source of a live CMS page:

Screenshot 2019-09-23 at 18 14 42

- run the static build (just the build is enough - I don't think we need to push to S3 to test this) eg, assuming you've set the hostname to be `example.net`:
```
docker-compose exec -e EXPORTED_SITE_URL="https://example.net" app python manage.py build --settings=developerportal.settings.production -v3
```
- shell into the `app` Docker container 
- cd into `build` 
- inspect the contents of the `index.html` for the page with a card that you edited (the directory structure is the same as the URL path). Where in the CMS Live page you'd have seen `localhost` for the absolute URLs for the `og` tags, you will see the hostname you've configured for the main Site. The hyperlinks between baked-out pages will be relative URLs:

eg:

Screenshot 2019-09-23 at 18 14 30

### DIFF
--- a/developerportal/settings/base.py
+++ b/developerportal/settings/base.py
@@ -212,12 +212,23 @@ WAGTAILIMAGES_IMAGE_MODEL = "mozimages.MozImage"
 # e.g. in notification emails. Don't include '/admin' or a trailing slash
 BASE_URL = os.environ.get("BASE_URL")
 
+
 # Wagtail Bakery Settings
 BUILD_DIR = os.path.join(BASE_DIR, "build")
 BAKERY_MULTISITE = True
 BAKERY_VIEWS = ("wagtailbakery.views.AllPublishedPagesView",)
 AWS_REGION = os.environ.get("AWS_REGION")
 AWS_BUCKET_NAME = os.environ.get("AWS_BUCKET_NAME")
+
+# Explicit configuration of where the 'baked' site will end up. This needs to match
+# the root URL of the developerportal Site in Wagtail's configuration, because
+# THAT value (Site.hostname) is what determines the domain used in any absolute
+# URLs generated.
+# However, we need to also have that root URL available here, because we must add it to
+# ALLOWED_HOSTS, else we will hit `DisallowedHost:Invalid HTTP_HOST header` during
+# baking in production mode, which then outputs all pages saying "400 Bad Request".
+
+EXPORTED_SITE_URL = os.environ.get("EXPORTED_SITE_URL")  # eg https://example.net
 
 # Static build management commands called in order
 STATIC_BUILD_PIPELINE = (("Build", "build"), ("Publish", "publish"))
--- a/developerportal/settings/production.py
+++ b/developerportal/settings/production.py
@@ -1,12 +1,24 @@
 from urllib.parse import urlparse
 
-from .base import *
+from .base import *  # noqa
 
 DEBUG = False
 DEBUG_PROPAGATE_EXCEPTIONS = True
-ALLOWED_HOSTS = [urlparse(BASE_URL).hostname] if BASE_URL else []
+
+ALLOWED_HOSTS = []
+
+if BASE_URL:
+    # This is the URL that Wagtail's CMS runs on.
+    ALLOWED_HOSTS.append(urlparse(BASE_URL).hostname)
+
+if EXPORTED_SITE_URL:
+    # This is the URL the static/baked site will be served from.
+    # It is different from the BASE_URL (where Wagtail is).
+    # We need it in ALLOWED_HOSTS to avoid `DisallowedHost`
+    # during baking
+    ALLOWED_HOSTS.append(urlparse(EXPORTED_SITE_URL).hostname)
 
 try:
-    from .local import *
+    from .local import *  # noqa
 except ImportError:
     pass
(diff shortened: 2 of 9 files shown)


stevejalim:
@dcgauld @liamnewmarch heya guys - if there was a reason that I've missed that you went down the path of using `request.site.root_url` in templates, please do shout. I _think_ this tweak doesn't have a downside, but never say never, etc. Cheers!

liamnewmarch:
@stevejalim this looks great, I’m in favour of using relative paths if only to save bytes!

The only problem I can think of is this might not work correctly if the live site is deployed to a subpath, e.g. to `example.com/portal/`, because _I think_ that path would be part of the `root_url`. This was the plan at one point but might have changed since we last talked about the prod URL.

stevejalim:
Hi @liamnewmarch 

> The only problem I can think of is this might not work correctly if the live site is deployed to a subpath

Yep, I agree. However, talking with @escattone the other day, it sounds like the decision was taken at the Mozilla side to elevate the portal to being on its own subdomain, rather than as a subpath.

With that in mind, are you happy to approve?

escattone:
> However, talking with @escattone the other day, it sounds like the decision was taken at the Mozilla side to elevate the portal to being on its own subdomain, rather than as a subpath.

Yes, that is the case. We've abandoned the idea of using a subpath of the MDN domain. The decision was made to publish to a separate domain.

stevejalim:
Argh, yes - that makes perfect sense: I just remembered I had an override
for `ALLOWED_HOSTS` lurking in `settings/local.py`. 🤦

Will backport the extra setting from #195, yep.

Thanks for the catch!

stevejalim:
@escattone @liamnewmarch Have backported the changes from closed PR #195 because - as Ryan correctly spotted - we do need to insert hostname for the target site into ALLOWED_HOSTS else baking will fail. 

(I missed this during my refactor because I had it specified in `settings.local`'s `ALLOWED_HOSTS`, but local.py has now been thrown in the woodchipper, never to return. I'll put $5 in the charity jar for my 🤦‍♂ )

Could you re-review, please? I've updated the PR description and the steps to test.

escattone:
> I'll put $5 in the charity jar for my 🤦‍♂
😆

escattone:
Looks great, thanks again @stevejalim !

escattone:
@stevejalim Added myself as an assignee simply for work-tracking purposes for zenhub

Issue #152, opened 2019-09-19
SVGs in HTML are too large


peterbe:
Large SVGs cause the HTML page to be very large and slow to download. The advantage of SVGs is that they can be controlled from the stylesheet, unlike binary images which are just static blobs. However, binary images load extremely fast comparatively. And images can be downloaded in parallel and cached by the browser too. 

```
> svgs= [...document.querySelectorAll('svg')]
(14) [svg, svg, svg, svg, svg, svg, svg, svg, svg, svg, svg, svg, svg, svg]
> svgs.map(o => o.outerHTML.length)
(14) [859, 859, 61869, 800, 1244, 2411, 1231, 1005, 1502, 1299, 5647, 14242, 1206, 61869]
> svgs.map(o => o.outerHTML.length).reduce((a, b) => a + b)
156043
> svgs.map(o => o.outerHTML.length).reduce((a, b) => a + b)  / 1024
152.3857421875
```

So, the HTML is weighed down by 152KB of SVG content.

PR #209, opened 2019-09-24
152: Reduce SVG sizes

stevejalim:
This changeset slims down three SVG files, to reduce page weight (as mentioned in #152) 

The optimisation was done with https://jakearchibald.github.io/svgomg/

Original SVG weight in an example page (featuring pattern and topic icons multuple times): 153KB
Post-optimisation of icon and panel background: 54KB

**Before**
Screenshot 2019-09-24 at 15 02 24

**After**
Screenshot 2019-09-24 at 15 02 37

The header SVG has also been optimised from 80KB down to 56KB.

**Before**

Screenshot 2019-09-24 at 15 23 19

**After**
Screenshot 2019-09-24 at 15 22 57

(If merged, should resolve issue #152)
## How to test

- pull branch
- view site
- be super-picky and look for imperfect results compared to before you pulled 😄


stevejalim:
@dcgauld I've only checked the SVG rendering in Firefox - let me know if it needs checking in other browsers and I'll do that, no worries.

dcgauld:
Looks great to me! Not sure how these were missed in the first place, but appreciate it 👍

Issue #235, opened 2019-09-26
Enable hourly rebuilds of whole site


stevejalim:
UPDATED DESCRIPTION:

A PR was made for this (https://github.com/mdn/developer-portal/pull/242) but was closed because it was agreed this should be triggered at the k8s level, not a cron in the `app` container.

Once https://github.com/mdn/developer-portal/pull/361 is merged, this ticket becomes unblocked.

We need:

* An hourly cron-like job that, hourly, runs `manage.py publish_scheduled_pages` (which means we can hide pages until we want them to appear in the site - see https://docs.wagtail.io/en/v2.6.2/reference/management_commands.html) 
* An hourly cron-like job that runs hourly but 15 mins after the previous command (to allow time for scheduled publishing to complete). This should call `manage.py enqueue_build_and_sync_task` (which re-exports the latest state of the site to S3 - see https://github.com/mdn/developer-portal/pull/242)

Note that we can't rely on just the first management command as a way to hourly re-export the site, because if there are no scheduled pages to be published, this won't have the side effect of triggering the build-and-sync process.

stevejalim:
@limed No rush, but is this something you could set up for us re Stage, please? (or feel free to point me at a way I can do it, if it's within my reach). Cheers!

stevejalim:
Ref https://github.com/mdn/developer-portal/pull/32

stevejalim:
Right now, no pages appear to support go-live/take-down dates in their Settings, so that would need adding too else the cronjob is pointless

stevejalim:
Expanding this to also reinstate scheduled baking and publishing to S3

stevejalim:
Update: this looks like it's better addressed at an ops level than the Draft PR I've opened - am hoping @limed has some time available, but it's not super-critical right now

stevejalim:
@limed @escattone I know we've agreed that an actual `crontab` entry in the `app` isn't the way forward, but now we're using a Celery worker, am thinking that `celery` in `beat` mode might be a decent way to kick of repeated things with less ops pain. (This is related to some task-queue-thinning work I'm doing right now where I think i'll need _another_ scheduled/cron task setting up)

stevejalim:
Once PR #412 is merged, I can do this as a simple extra celerybeat change

PR #242, opened 2019-09-26
235: Add and reinstate cron jobs

stevejalim:
This changeset addresses the need to auto-publish the site to S3 regularly, as well as the standard Wagtail need to support scheduled publishing/unpublishing of pages (#235)

Notes/context: 

* While we won't be automatically un-publishing Event pages once they have passed, we may want to auto-publish them when the time is right. 
* Events that are in the past will be accessible via their regular URIs, but they should not appear in lists of upcoming events for more than a day after they have taken place. The code that renders such a list will use Django's `filter()` method to not show them in the site. As such, we need to regularly bake out a new static site to ensure that the static content is as up to date as possible.

## Key changes:

* Add job to trigger scheduled publishing of pages
* Reinstate job to build a new static site and sync it to S3

Note these changes are based closely on what was done before and then removed and those changes came with some comments that sound like there may be a better way (hence this only being a draft PR right now). @escattone - can you clue me in to what you had in mind when you said `"Note-to-self: Modify to use our build script."` re these two adjacent comments, please?

- https://github.com/mdn/developer-portal/pull/29#discussion_r305553372
- https://github.com/mdn/developer-portal/pull/29#discussion_r305553517 

## How to test

- locally, `docker-compose build` to run the setup again:
```
...
Step 12/25 : RUN crontab -l | { cat; echo '0 * * * * /usr/local/bin/python /app/manage.py publish_scheduled_pages'; } | crontab -
 ---> Running in ec2c0ca84d3f
Removing intermediate container ec2c0ca84d3f
 ---> 40e5dbf45472
Step 13/25 : RUN crontab -l | { cat; echo '10 * * * * /usr/local/bin/python /app/manage.py build && /usr/local/bin/python /app/manage.py publish'; } | crontab -
 ---> Running in 5eef8fa87706
Removing intermediate container 5eef8fa87706
...
```

Shell into the container and check the crontab:
```
/app # crontab -l
# do daily/weekly/monthly maintenance
# min   hour    day     month   weekday command
*/15    *       *       *       *       run-parts /etc/periodic/15min
0       *       *       *       *       run-parts /etc/periodic/hourly
0       2       *       *       *       run-parts /etc/periodic/daily
0       3       *       *       6       run-parts /etc/periodic/weekly
0       5       1       *       *       run-parts /etc/periodic/monthly

0 * * * * /usr/local/bin/python /app/manage.py publish_scheduled_pages
10 * * * * /usr/local/bin/python /app/manage.py build && /usr/local/bin/python /app/manage.py publish
/app # 
```

## Thoughts / open questions

* Is there a better way than this?  Happy to change
* If this now runs in local dockers, do we want to add an env-var flag to disable it locally?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,12 @@ FROM python:3.7-alpine@sha256:488bfa82d8ac22f1ed9f1d4297613a920bf14913adb98a652a
 EXPOSE 8000
 WORKDIR /app/
 
+# Update cron to publish/unpublished scheduled pages hourly
+RUN crontab -l | { cat; echo '0 * * * * /usr/local/bin/python /app/manage.py publish_scheduled_pages'; } | crontab -
+
+# Update cron to "bake" a new static site each hour and push it to S3 - allows 10 mins for auto-publishing to complete
+RUN crontab -l | { cat; echo '10 * * * * /usr/local/bin/python /app/manage.py build && /usr/local/bin/python /app/manage.py publish'; } | crontab -
+
 RUN apk add --no-cache --virtual .build-deps \
   gcc \
   musl-dev \
@@ -41,4 +47,5 @@ COPY --from=static /app/dist /app/dist/
 
 # Collect all of the static files into the static folder
 RUN DJANGO_ENV=production python manage.py collectstatic
-CMD exec gunicorn developerportal.wsgi:application --bind=0.0.0.0:8000 --reload --workers=3
+CMD crond -d 8 -L /var/log/cron.log && \
+  exec gunicorn developerportal.wsgi:application --bind=0.0.0.0:8000 --reload --workers=3


stevejalim:
@escattone - could you chip in on this when you get a mo, please, particularly ?
@limed I'd really appreciate your view, too - is there anything i've overlooked?

escattone:
@stevejalim Added myself as an assignee simply for work-tracking purposes for zenhub

escattone:
@stevejalim Regarding my two comments:

> Note-to-self: For our use case, I need to replace /usr/local/bin/python /app/manage.py build with our build script.

> Note-to-self: Modify to use our build script.


You can completely ignore them! 😄 They were made within the context of the `build` target being a sub-path of the MDN domain (e.g., https://developer.mozilla.org/portal), and in that case I had created a separate (and half-finished) `build-portal` command to handle the non-standard "needs" of that case (see https://github.com/mdn/developer-portal/pull/33 and https://github.com/mdn/developer-portal/pull/88 for history).

stevejalim:
@escattone Thanks for the feedback, and I agree that making it a k8s `CronJob` is better - it's more visible and cleaner (so gets rid of my "we want to stop this running in local docker" issue etc)

@limed Would you be OK to take over this ticket and implement the crons that are specced here, please? Feel free to do it on a separate branch and close this draft PR unmerged. Cheers!

limed:
> @limed Would you be OK to take over this ticket and implement the crons that are specced here, please? Feel free to do it on a separate branch and close this draft PR unmerged. Cheers!

Yup I can deal with the `CronJob` but I believe we should wait for the celery work to be done first?

stevejalim:
@limed This isn't blocking anything, no, so waiting for the Celery branch
to be merged is fine.

Indeed, thinking about it, once we have the Celery branch merged, that is
gonna change the chain of actions, potentially.

I'll write it up in a doc/wiki page so we can all get in the same headspace
about how it should roll. Suggestions will be very welcome!

stevejalim:
@limed See https://github.com/mdn/developer-portal/pull/280 -- got a couple of questions on there that I'd apprecaite your input on

stevejalim:
@limed The spec for the jobs to be run is going to change slightly to use a new management command instead of the ones mentioned in the current diff - see https://github.com/mdn/developer-portal/issues/296

stevejalim:
Closing because this should be done at the k8s level, as per discussion above

PR #249, opened 2019-09-27
Ensure draft pages aren’t pulled through

### LINKED ISSUE #243: Ensure draft pages aren’t pulled through
There are a few cards on the homepage of the staging site, and on other pages, which lead to a 404 page. This is because they are cards for draft articles which are mistakenly being pulled through.

Most of our queries include `.live().public()` as per [this note][1] in the Wagtail docs which prevents listing draft/private pages, but some do not.

[1]: http://docs.wagtail.io/en/v2.0/reference/pages/queryset_reference.html#wagtail.core.query.PageQuerySet.public

liamnewmarch:
This changeset addresses the issue of draft resources being pulled through in various places by providing:

* An additional `published_objects` manager for Page-based models, and
* A `published` filter for use with StreamField properties (which can contain a mix of Page and non-Page items).

(Resolves #243)

## Key changes:

* Add abstract BasePage class with published_objects manager

* Use published_objects for resource queries

* Add `published` filter for StreamField properties

* Fix isort spacing issue


developerportal/apps/articles/migrations/0032_auto_20190930_0915.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 2.2.4 on 2019-09-30 09:15
```
stevejalim: Nitpick: don't worry about it here, but for future migrations, I find it's really useful to rename them to something explicit (eg `0032_update_article_manager.py`) - that makes them easier to grok when looking at `showmigrations`, etc

developerportal/templatetags/app_filters.py
```diff
@@ -24,6 +26,14 @@ def hex_to_rgb(hex_color, alpha=1):
     return f"rgb({r}, {g}, {b})" if a == 1 else f"rgb({r}, {g}, {b}, {a})"
 
 
+@register.filter(name="published")
+def published(items):
+    """Filters StreamField items to remove drafts Pages."""
```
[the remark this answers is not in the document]
stevejalim: I appreciate this might be a pain to test in the context of a real StreamField, but any chance of a unit test for this (using simpler data structures for `item` perhaps)?

developerportal/apps/common/models.py
```diff
@@ -0,0 +1,19 @@
...
+class PublishedPageManager(PageManager):
+    def get_queryset(self):
+        # This method restricts the query to only live (non-draft) pages - it
+        # doesn't guard against private pages, however these aren't supported
+        # by the static site build.
+        return super().get_queryset().live()
+
+
+class BasePage(Page):
+    base_form_class = BasePageForm
```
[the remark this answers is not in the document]
liamnewmarch: Thanks for picking up on this, it’s I meant to do and clearly forgot.

developerportal/apps/articles/migrations/0032_auto_20190930_0915.py
liamnewmarch: Good point, it’s a bad habit I/we’ve fallen into. I’ll rename these before they land in master and will make sure to name them in the future.

developerportal/apps/common/models.py
```diff
@@ -0,0 +1,19 @@
...
+
+from .forms import BasePageForm
+
+
+class PublishedPageManager(PageManager):
+    def get_queryset(self):
+        # This method restricts the query to only live (non-draft) pages - it
+        # doesn't guard against private pages, however these aren't supported
+        # by the static site build.
+        return super().get_queryset().live()
```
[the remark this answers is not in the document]
liamnewmarch: That’s right. Specifically, the reason is that adding `.public()` here breaks running migrations because it tries to look up the `wagtailcore_pagerevision` table before it’s been created (hence Travis CI failing earlier). I’ll clarify this in the code comment.

developerportal/apps/events/models.py
```diff
@@ -13,6 +13,7 @@
 )
 from django.utils.safestring import mark_safe
 
+from django_countries.fields import CountryField
```
stevejalim: How come this is added during this PR? Was it absent before and only working because of a lucky import elsewhere?

developerportal/apps/externalcontent/models.py
```diff
@@ -10,6 +10,7 @@
     URLField,
 )
 
+from django_countries.fields import CountryField
```
stevejalim: And again - backfilling a missing import, yeah?

developerportal/apps/people/models.py
```diff
@@ -263,18 +258,16 @@ def events(self):
...
-        upcoming_events = (
-            Event.objects.filter(start_date__gte=datetime.datetime.now())
-            .live()
-            .public()
+        upcoming_events = Event.published_objects.filter(
+            start_date__gte=datetime.datetime.now()
         )
 
-        speaker_events = Event.objects.none()
+        speaker_events = Event.published_objects.none()
```
stevejalim: Tangent: what's the use-case for `none()` here?

developerportal/apps/events/models.py
liamnewmarch: This import was grouped incorrectly by the author and isort picked it up (see line 33)

developerportal/apps/people/models.py
liamnewmarch: I believe the author’s intent here is to create an empty queryset and populate it with the subsequent loop. These methods can probably be refactored to be more efficient, I’ll create an issue.

developerportal/apps/events/models.py
stevejalim: 🤦‍♂ Thanks!

Issue #255, opened 2019-09-30
Ops: make Redis available in k8s deployment


stevejalim:
The changes required for https://github.com/mdn/developer-portal/issues/139 including adding Redis, because Celery no longer works with a postgres backend.

As such, merging the PR for that issue is blocked (or risks breaking staging, in real terms) until redis is available and its location is exposed via a `REDIS_URL` env var.

stevejalim:
Heya @limed - if you have time/headspace for this, that would be epic. It's not super-urgent, though as the Celery PR mentioned above can still be reviewed etc.

limed:
@stevejalim see #261 I believe that even exposing the environment variable the app still needs a celery worker? If so let me know

limed:
See https://github.com/mdn/infra/pull/305 where the redis cluster was created in AWS

stevejalim:
> limed: I believe that even exposing the environment variable the app still needs a celery worker? If so let me know

@limed yeah you're right. Having redis available is only half the deal, but it at least lets Django put Celery jobs into redis and not `500` because it can't find Redis. 

The next step is, indeed, getting k8s to run a Celery worker process. It's going to probably have to come after https://github.com/mdn/developer-portal/pull/256 is merged, else the worker will die. I'll make a separate ticket for it now

stevejalim:
That separate ticket is https://github.com/mdn/developer-portal/issues/267

Issue #252, opened 2019-09-28
Consider adding link to live static site in flashes


stevejalim:
When a page is published, wagtail shows a link to view the live page. There are also lots of links around to visit the live page. These links point to the CMS's live pages.

Now, however, we have a static live version too, should we offer links to that in addition? 

Is this actually a problem at all?

PR #256, opened 2019-09-30
139: Fix auto-creation of static build following publish/unpublish

### LINKED ISSUE #139: Check/fix automatic, post-publish static build
Currently, there are handler for the `published` and `unpublised` Signals, which then trigger a new `subprocess.Process` to build the static site.

Unforunately, they don't appear to be working locally because the DB becomes unavailable:

```
app_1     | INFO:developerportal.apps.staticbuild.wagtail_hooks:Static build task ‘Build’ started.
app_1     | INFO:bakery.management.commands.build:Build started
app_1     | Process Process-1:
app_1     | Traceback (most recent call last):
app_1     |   File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
app_1     |     return self.cursor.execute(sql, params)
app_1     | psycopg2.OperationalError: server closed the connection unexpectedly
app_1     | 	This probably means the server terminated abnormally
app_1     | 	before or while processing the request.`
```

- [ ] tail the postgres logs when it happens - is there something schema-related that's killing the DB?
- [ ] consider moving to a different approach (eg task queue with `celery`/`django-rq`) if `Process` can't be made to reliably work

(Note that building the site via management command from the CLI seems to work fine)

stevejalim:
This changeset ensures that an automatic build-and-sync-to-S3 of the site takes place when a page is published or unpublished.

This approach uses a task queue (Celery) rather than spawning a new `subprocess.Process`. This should be more stable and visible.

(Resolves #139)

## Key changes:

* Update Dockerfile to:
  - rename `app` target as `app_base`, to make it clearer it's being used for multiple things
  - not start gunicorn as part of `app_base`'s build, because that's not needed and is handled by docker-compose or k8s

* Update docker-compose.yml to:
  - add `redis` container
  - update the `app` config to inherit from a new `common` config, which uses the `app_base` target
  - add new `worker` container, inheriting from the `common` config, too

* Add `celery` to the project
  - Stop the use of `subprocess.Process` when a page is published or unpublished.
  - Use a a properly enqueued task to handle the background dump-to-static-then-upload-to-S3 process

*  Add `django-celery-results` to give us easier visibility of job success or failure

    This is a temporary and basic solution to give us an eye on whether background tasks are
happy, but it is limited because it only displays tasks after they have completed (ie
once they have a result).
    Switching to something that lets us see in-flight tasks would be useful, down the line. (eg [Flower](https://flower.readthedocs.io/en/latest/features.html), though that's a sledgehammer to crack a nut)

*  Ensure Django (and therefore the Celery worker) can get hold of the credentials Celery needs

  - Update Makefile to export AWS-related config to the environment
  - In Django, pick up AWS-related config from environment

    This change assumes the AWS-related config is already available at in the general config
at the k8s level -- this seems reasonable, because they are all already mentioned in the
k8s config files.

*  Add light integration tests to show that publishing and unpublishing will trigger static-build-and-sync behaviour

## Things to note

* The standard pipeline of `manage.py build` and `manage.py publish` results in **ALL CMS pages always being uploaded, even if they are not directly affected by any changes**. However, only changed staticfiles are uploaded each time. For now, the 'all the pages' approach should be OK, but I expect we'll need to revisit that once the amount of content increases.

* We need to do a bit of thinking about how this background publishing approach will feel in reality:
    * Related to https://github.com/mdn/developer-portal/issues/252, the UI currently makes no mention of the static site, and therefore no mention that the static site will take X seconds to be baked and synced to S3. This might be something we can solve with documentation/editor education, but it would be good to come up with a more robust solution
   * django-celery-results shows us when build-and-sync has completed, but not when it's in progress - Flower should help there
   * We do need to enable hourly scheduled publishing, so that things like events listings naturally decay/update over the course of a day, rather than waiting for the next time an editor publishes content. However, there's a (minor?) risk of race conditions causing confusing output if - essentially - two separate tasks are publishing the site at the same time. If this is a problem, we can look to get around it by checking what's in Celery's Redis before deciding whether to do a build-and-sync, or not. Up for debate, for sure!

* After running into a speedbump with Celery, I tried `dramatiq` based on @escattone's suggestion. However, `dramatiq` automatically tries to import any module called `tasks.py`, which causes `django-bakery` to explode, because django-bakery [_expects_ Celery to be available when its tasks module is imported](https://github.com/datadesk/django-bakery/blob/master/bakery/tasks.py#L6)...

* This could definitely do with some laser-eyes from @escattone and @limed re how it'll slot into our staging and production infra. I have opened a soft-blocking issue (https://github.com/mdn/developer-portal/issues/255) to add Redis, because Celery needs it.

* If you want to run the async build locally, you'll need to set the following in your `settings/local.py`:
    * `AWS_ACCESS_KEY_ID`
    * `AWS_SECRET_ACCESS_KEY`
    * `AWS_BUCKET_NAME`
    * `AWS_REGION`
and set `DEBUG = False`

* Note, too that the Celery `worker` will need manually restarting to pick up code changes. It doesn't auto-detect them like `runserver` does. 

## How to test

- check out this branch
- run migrations (for `django-celery-results`' tables)
- add/update your `settings.local` to include the four AWS settings mentioned above, as well as turning DEBUG mode off. (Make sure you don't use any production-type AWS credentials for this testing!!) 
- ensure the target S3 bucket exists
- boot up the stack with `docker-compose build && docker-compose up`
   - if you get problems, going nuclear with `docker-compose down --volumes && docker-compose build && docker-compose up` and/or using the `dev-setup --purge` should rebuild all you need. You may need to create a new Django/Wagtail superuser to log in to the CMS
- Pick a page and publish it.
- Check the console to see `worker_1` baking a static site then syncing it to S3
- Check the S3 bucket to see the site contents all there. Download the bucket and view in a local browser if you want.
- Remove the baked site from your S3 bucket


k8s/Makefile
```diff
@@ -5,6 +5,10 @@ export AWS_RESOURCE_STACK=dev-portal-${TARGET_ENVIRONMENT}
 export AWS_REGION ?= us-west-2
 export K8S_CLUSTER_SHORT_NAME ?= eks-oregon
 
+export AWS_BUCKET_NAME = ${AWS_BUCKET_NAME}
+export AWS_ACCESS_KEY_ID = ${AWS_ACCESS_KEY_ID}
+export AWS_SECRET_ACCESS_KEY = ${AWS_SECRET_ACCESS_KEY}
+
```
stevejalim: @limed Can you cast some laser eyes over this and the Docker changes, please? Am pretty sure the vars mentioned above all exist already in k8s, yep?

Dockerfile
```diff
@@ -41,4 +41,6 @@ COPY --from=static /app/dist /app/dist/
 
 # Collect all of the static files into the static folder
 RUN DJANGO_ENV=production python manage.py collectstatic
-CMD exec gunicorn developerportal.wsgi:application --bind=0.0.0.0:8000 --reload --workers=3
```
stevejalim: I checked with @limed and this line looks redundant - docker-compose also runs it, and k8s also has a version of it

docker-compose.yml
```diff
@@ -10,24 +10,39 @@ services:
...
+  app:
+    <<: *common
+    command: gunicorn developerportal.wsgi:application --bind=0.0.0.0:8000 --reload --workers=3
+    ports:
+      - '8000:8000'
+      - '8080:8080'
+
+  worker:
+    <<: *common
+    command: celery -A developerportal.apps.staticbuild worker -l info
```
stevejalim: The only niggle I can see with running celery like this is the warning

`RuntimeWarning: You're running the worker with superuser privileges: this is absolutely not recommended!`

@escattone / @limed Do either of you have a suggestion of how to tackle this, please, or if we def need to (given we don't use this docker-compose.yml in k8s, yep?)

k8s/Makefile
limed: Yes this variables already exists in the k8s manifest so you shouldn't need to worry about doing this here

Dockerfile
limed: :+1: here the only time this is really only useful is if you are running this using the `docker run` command. Since we are using lots of `docker-compose` this is really not needed

docker-compose.yml
```diff
@@ -10,24 +10,39 @@ services:
     volumes:
       - database:/var/lib/postgresql/data
 
-  app:
+  redis:
+    env_file: .env
+    expose:
+      - 6379
+    image: redis:5.0.6-alpine
```
limed: Not sure if it makes a difference but AWS redis engine I believe only goes up to `5.0.5` so maybe we can change the image name to match that.

k8s/Makefile
limed: Yes these already exists in k8s you don't even need this here I believe

Dockerfile
stevejalim: @limed Great, thanks. Just so I get it right in my head, is Prod using _anything_ defined in `docker-compose` at all?

k8s/Makefile
stevejalim: OK will delete them. Cheers!

docker-compose.yml
stevejalim: Ah, Docker Hub doesn't have an official 5.0.5 redis: https://hub.docker.com/_/redis/

(There is one around, but it's of unknown provenance https://hub.docker.com/search?q=redis%3A5.0.5&type=image)

OK to stick with 5.0.6?

Dockerfile
limed: No nothing in k8s will be using `docker-compose`. See: https://github.com/mdn/developer-portal/blob/master/k8s/app.deploy.yaml.j2#L29

developerportal/apps/staticbuild/celery.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+from django.conf import settings
+
+from celery import Celery
```
liamnewmarch: I think `celery` will need to be added to **.isort.cfg** too.

docker-compose.yml
liamnewmarch: Docker recommend [running services as a non-privileged user][1] if possible. It sounds like that would fix the warning but I’m not sure what else it would impact.

[1]: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user

stevejalim:
@limed and/or @escattone + @liamnewmarch Could you please take a look at my changes to the Dockerfile and docker-compose.yml in  5a790c7 please? They were done following Celery's warning about running as root in a container (👍 to Celery!), but I'd appreciate scrutiny from y'all re whether it's a safe change (partic re the new user's creation).

Cheers! 
Steve

stevejalim:
Addressed various comments. Good to merge. Got a 👍 in Slack from @limed

PR #263, opened 2019-10-01
222: Add default image to meta and fix other properties

### LINKED ISSUE #222: Add default image for OG image link fallback
The Open Graph meta tags currently feature an image link where possible, using either the Card image if configured, falling back to the Page's image, if configured. 

However, there's no image shown if neither of the above is configured.

Might be good to have a sensible default image.

dcgauld:
Currently blocked on #249.

This changeset adds a default social image to meta tags, and ensures that relevant meta properties are pulled through from Wagtail fields correctly.

(Resolves #222)

## Key changes:

- Add a placeholder image when no main or card image is uploaded.
- Ensures a valid description is pulled through from pages where possible (including article intros).
- Pull through keywords correctly (not used much nowadays but still potentially useful).


developerportal/templates/atoms/social-meta.html
```diff
@@ -1,31 +1,37 @@
...
-    <meta property="og:image:width" content="{{ image.width }}">
-    <meta property="og:image:height" content="{{ image.height }}">
-  {% endif %}
+{% static "img/placeholders/article.jpg" as fallback_image_url %}
+{% if page.card_image %}
+  {% image page.card_image width-1200 as image %}
+{% elif page.image %}
+  {% image page.image width-1200 as image %}
 {% endif %}
+<meta property="og:image" content="{{ base_url }}{% if image %}{{ image.url }}{% else %}{{ fallback_image_url }}{% endif %}">
```
stevejalim: Nit: could we use the `firstof` tag here instead of the `if...else`?

developerportal/templates/atoms/social-meta.html
```diff
@@ -1,31 +1,37 @@
...
+<meta property="og:image:width" content="{% if image %}{{ image.width }}{% else %}960{% endif %}">
```
stevejalim: Super nitpick: if the article.jpg placeholder is recut or replaced at any point, these would become invalid. Maybe defining them along with the `fallback_image_url` a few lines above would at least put all the things that might need to be tweaked adjacent to each other. 

See what I mean about super-nit-picky? 😄 Feel free to skip if you disagree

developerportal/templates/atoms/social-meta.html
dcgauld: Good catch, thanks!

developerportal/templates/atoms/social-meta.html
dcgauld: I've added some variables for the width/height with `{% with %}`. Not sure if this is what you were meaning, but let me know!

PR #268, opened 2019-10-01
137: Prevent past events from showing in resource lists

### LINKED ISSUE #137: Ensure that past events are not displayed
## Current state

Stakeholder comments say that 
> Events will not discoverable after the event date has passed

However, there doesn't yet appear to be support for this.

## Suggested solution

* Add support for only showing today-and-future events. 
  * Server time is UTC, which presents a challenge. We could hack around it and use JS to set a cookie value, which we read server-side, but that's messy and not robust (eg, won't work if a user goes straight to the page, so there's no opportunity to trigger the JS to first stash the TZ in the cookie before needing to render the page which would need to know the TZ).
  * Simplest fix might just be to show 'yesterday-to-future' events -- stakeholder confirmation needed.

dcgauld:
This changeset prevents past events from being shown within the events resource lists.

(Resolves #137)

## Key changes:

- Add filter to only return future events within resource lists.
- Move "no resource" logic into filter list component, and add specific message support.

## How to test

- Add events, one future and one past, and check that only the future one is shown in the event list.
- Change the future event's start date to the past and make sure the events directory is empty.

### DIFF
--- a/developerportal/apps/events/models.py
+++ b/developerportal/apps/events/models.py
@@ -139,7 +139,7 @@ class Events(Page):
     @property
     def events(self):
         """Return events in chronological order"""
-        return get_combined_events(self)
+        return get_combined_events(self, start_date__gte=datetime.date.today())
 
     def get_filters(self):
         from ..topics.models import Topic
@@ -330,7 +330,7 @@ class Event(Page):
     @property
     def is_upcoming(self):
         """Returns whether an event is in the future."""
-        return self.start_date > datetime.date.today()
+        return self.start_date >= datetime.date.today()
 
     @property
     def primary_topic(self):
--- a/developerportal/templates/organisms/filter-list.html
+++ b/developerportal/templates/organisms/filter-list.html
@@ -4,39 +4,45 @@
       {% include "atoms/pattern.html" %}
     </div>
     <div class="container">
-      <div class="filter-list-content">
-        <aside class="filter-list-sidebar">
-          <details class="filter-list-sidebar-content filter-list-sidebar-content-mobile">
-            <summary>Filter</summary>
-            {% include "molecules/filter-form.html" with resources=resources filters=filters initial_resources=initial_resources resources_per_page=resources_per_page %}
-          </details>
-          <div class="filter-list-sidebar-content filter-list-sidebar-content-desktop">
-            {% include "molecules/filter-form.html" with resources=resources filters=filters initial_resources=initial_resources resources_per_page=resources_per_page %}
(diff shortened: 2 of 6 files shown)


developerportal/apps/events/models.py
```diff
@@ -330,7 +330,7 @@ class Event(Page):
     @property
     def is_upcoming(self):
         """Returns whether an event is in the future."""
-        return self.start_date > datetime.date.today()
+        return self.start_date >= datetime.date.today()
```
liamnewmarch: This makes sense to stay as `start_date` (as opposed to the filtering above) but, being really pedantic, I think we should go back to `>`.

developerportal/apps/events/models.py
```diff
@@ -139,7 +139,7 @@ def get_context(self, request):
     @property
     def events(self):
         """Return events in chronological order"""
-        return get_combined_events(self)
+        return get_combined_events(self, start_date__gte=datetime.date.today())
```
liamnewmarch: Now that I think about it, maybe we should be filtering on `end_date` here? i.e. filtering out events which have finished, rather than those which have started.

developerportal/templates/organisms/filter-list.html
```diff
@@ -4,39 +4,45 @@
...
+            </div>
+            <div class="filter-list-actions" id="js-filter-list-actions">
+              <button class="mzp-c-button" id="js-filter-list-action-next-page" type="button">See more</button>
+            </div>
           </div>
         </div>
-      </div>
+      {% else %}
+        <div class="filter-list-no-results">
+          <h2>{{ no_resources_message|default:"No resources found" }}</h2>
```
liamnewmarch: Nice, I like this pattern.

developerportal/apps/events/models.py
stevejalim: I get what you're saying, but I think we can live with `>=` given that "today" is quite a bunch of waking hours across the globe, yeah?

developerportal/apps/events/models.py
stevejalim: Is now a decent time to slot in some tests for these methods?

dcgauld: Yeah, great points. I'll get on these in a bit!

liamnewmarch: We’ve agreed to stick with `start_date` for now and raise a ticket to use `end_date` and possibly make it required (discussion needed?) I’m writing tests for this method now, thanks!

PR #269, opened 2019-10-01
Use correct check for icons within featured items

dcgauld:
This changeset fixes an incorrect external item check for featured card icons.


dcgauld:
Good shout, will do. Thanks!

Issue #267, opened 2019-10-01
Ops: add Celery worker to consume static-build tasks


stevejalim:
When merged, https://github.com/mdn/developer-portal/pull/256 will support for async baking and syncing of the site to S3, via Celery.

Celery needs Redis, which has been added in PR https://github.com/mdn/developer-portal/pull/261 so it will be able to put jobs into the queue.

What we need as a follow-up (and this is maybe until #256 is merged, because it adds `celery` and a specific `worker` container, for instance) is a worker process that consumes the queued Celery tasks. The `docker-compose.yml` in #256 shows how to invoke it. 

Please do ping me if you have any questions!

PR #277, opened 2019-10-02
Added celery worker

limed:
Adds celery worker pods

(Resolves part of #267)

## Key changes:

- Adding celery worker pods


limed:
I accidently broke this PR so I'm resubmitting

PR #280, opened 2019-10-02
Add initial doc explaining how the static site will be generated.

stevejalim:
There are still things that are TBD, but this gets it moving.

### DIFF
--- a/docs/automatic-publishing-to-s3.md
+++ b/docs/automatic-publishing-to-s3.md
@@ -1,0 +1,27 @@
+# Automatic publishing of a static site to S3
+
+[Wagtail Bakery](https://github.com/wagtail/wagtail-bakery) can build ("bake") a static version of the site and sync it to Amazon S3, where it can be served as a website.
+
+_Note that "Publish" and "Unpublish" here refer to the Wagtail sense of making a page live / not live. Emitting a version of the site to S3 will be referred to as "build-and-sync"._
+
+In production, this "build-and-sync" process will run at a few points:
+
+1. It will be run whenever a Page is Published or Unpublished from the Wagtail admin. (Note: this means if two Pages are published, the "build-and-sync" process is run twice...)
+
+2. Hourly, on a schedule, we'll check for any Wagtail Pages due to be automaticaly Published or Unpublished, and if any are found, that will be done. This action will then trigger the "build-and-sync" step mentioned in 1 for each page being automatically Published or Unpublished.
+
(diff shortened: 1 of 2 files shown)


docs/automatic-publishing-to-s3.md
```diff
@@ -0,0 +1,34 @@
...
+1. Schedule a job that runs in the main app container and executes the two relevant wagtail-bakery commands:
+
+   ```
+   /usr/local/bin/python /app/manage.py build &&
+   /usr/local/bin/python /app/manage.py publish
+   ```
+
+This will be synchronous, but should not interfere with the app's request-response cycle. We could probably also do this in the `worker` container.
+
+2. We add a new management command that emits the post-Publish Django Signal that starts **Case 1**, so we end up with a new queued task, which the worker picks up.
```
stevejalim: This seems cleaner to me because it puts all the S3 action into worker processes.

That said, it'll be harder to spot failing scheduled jobs because there won't be much to fail other than "add a new task to the queue"

Opinions welcome!!

docs/automatic-publishing-to-s3.md
```diff
@@ -0,0 +1,34 @@
...
+
+## A note about queue load
+
+There is a risk that the various ways of triggering an async build, plus the fact that each Publish/Unpublish action for each page will trigger a new static build, means there may be a lot of "build-and-sync" tasks in the queue, and all but the most recent one will be redundant. (We might want to look at a way to purge outmoded "build-and-sync" tasks and just use the newest one.)
```
stevejalim: > We might want to look at a way to purge outmoded "build-and-sync" tasks and just use the newest one.

I think this is the way to go -- what does everyone else think?

stevejalim:
Feedback and suggestions are really welcome here -- there may be simpler or cleaner solutions to be had!

limed:
The only thing I might add here is during the unpublish process we should consider doing something about invalidating the cache from cloudfront.

stevejalim:
@limed great call. Let's lean on Wagtail for FE cache invalidation: https://docs.wagtail.io/en/v2.6.2/reference/contrib/frontendcache.html#frontend-cache-purging

I'll ticket it up

limed:
Everything else looks good to me. r+

docs/automatic-publishing-to-s3.md
escattone: I think so too, that sounds like a nice feature to have!

tobinmori:
Late to the party - but looks good to me too. Nice call on cache invalidation.

PR #282, opened 2019-10-02
144: video modal

### LINKED ISSUE #144: Open external videos in a modal instead of linking out
We should be able to use [Protocol's modal component](https://protocol.mozilla.org/patterns/organisms/modal.html) for this. We may also need to limit the video providers that are supported (e.g. just support YouTube embeds for now).

liamnewmarch:
This changeset addresses opening external videos in a modal (YouTube only, for now).

(Resolves #144 )

## Key changes:

- Updates Protocol to the latest version (9.0.1) which necessitated a slight restructure to our **src/img** directory, hence the number of files changed.
- Implements a new Modal atom component which can be triggered from any link with `.js-modal-trigger` and a valid YouTube URL as its href.

## How to test

- Create a new external video with a YouTube URL and choose a topic.
- Navigate to the topic page and click the video’s card.


src/js/atoms/modal.js
```diff
@@ -0,0 +1,92 @@
+// Transforms a normal YouTube URL into an iframe embed URL
+function transformYouTubeURL(url) {
+  const re = /^.*(youtu.be\/|v\/|u\/\w\/|embed\/|watch\?v=|&v=)([^#&?]*).*/;
```
stevejalim: Does this regex catch video URLs that aren't on YT's "cute" domain?

liamnewmarch: It does because the `(youtu.be\/|v\/|u\/\w\/|embed\/|watch\?v=|&v=)` capture group is testing for strings that will directly precede the ID, but that does highlight a big problem with this regex – it’s way too permissive in what it accepts.

I’ve made some modifications to it:

```regex
/^https?:\/\/(?:(?:www\.)?youtube(?:-nocookie)?\.com.*(?:\/|v\/|u\/\w\/|embed\/|\?v=|&v=)|youtu.be\/)([\w-]{11})/
```

- String must start http:// or https://.
- Strict domain matching. Either youtube.com, youtube-nocookie.com – with or without www subdomain – or youtu.be.
- Explicit characters (`[\w-]`) and length `{11}` for the ID.
- Non-capturing groups throughout except the ID.

I’ve tried with a sample of YouTube IDs but will test against more before I commit it.

PR #316, opened 2019-10-07
Standardize CI init / setup scripts

### LINKED ISSUE #315: Standardize ci test setup
Currently we have travis ci running test agains every PR and commit to master, we do the same in jenkins as well. Unfortunately the way we setup the environments are different for both CI systems where 1 system uses the `docker-compose.test.yaml` file and the other uses a bunch of scripts in the `scripts` folder.

This causes confusion and we should merge them and have them just use one method instead of two different methods.

limed:
Remove the `docker-compose.test.yaml` file and standardize the way we setup the test environment for both jenkins and travis. Now everything should just run off the setup scripts in the `scripts` folder

(Resolves #315)

## Key changes:

- Remove reference to `docker-compose.test.yaml` in `Jenkinsfile`
-  Add an optional `--build` flag in `ci-setup`


Jenkinsfile
```diff
@@ -39,14 +39,13 @@ node {
...
       stage ('Test') {
-        withEnv(["COMPOSE_FILE=docker-compose.test.yml"]) {
-          sh "docker-compose up --detach"
-          try {
-              sh "docker-compose exec -T app ./manage.py migrate"
-              sh "docker-compose exec -T app ./manage.py test"
-          } finally {
-              sh "docker-compose down --volumes --remove-orphans"
-          }
+        sh "docker-compose up --detach"
```
escattone: I suspect this might not work, since the `docker-compose.yml` file doesn't know anything about the `mdnwebdocs/developer-portal:<tag>` images? I'm wondering if it might be best to delete this line, and then...

Jenkinsfile
```diff
@@ -39,14 +39,13 @@ node {
...
+        try {
+            sh "scripts/ci-setup"
```
escattone: ...change this to `sh "scripts/ci-setup --build"`, as well as...

Jenkinsfile
```diff
@@ -39,14 +39,13 @@ node {
...
+            // Consider running this through scripts/ci-tests ?
+            sh "docker-compose exec -T app ./manage.py test"
```
escattone: ...change this to `sh "scripts/ci-tests"`?

Jenkinsfile
limed: I was going under the assumption that it will grab the images it built, it just seems very inefficient if we just did a docker build previously and now we are running a `docker-compose up --build --detach`

Perhaps it should be a second pass optimization thing.

Jenkinsfile
limed: Agreed here, this also test npm though. But its not a big deal

Issue #318, opened 2019-10-07
Add CNAME (and/or CDN) for user-media in S3


stevejalim:
Once the bucket for #317 is set up, it'll probably need a decent public-friendly CNAME. 

It might well also benefit from a a CDN (eg Cloudfront), but there is a catch: invalidating it fom Wagtail won't be easy because Wagtail's multiple-frontend-cache-invalidation-code is written to invalidate pages across multiple **Sites**, and there won't be a Site that exists that matches the CNAME/hostname of this new CDN.

So, we might have to do some custom invalidation work if we do go with a CDN (see https://docs.wagtail.io/en/v2.0/reference/contrib/frontendcache.html#advanced-usage)

@limed - what do you reckon?

stevejalim:
Heya @limed - having CNAMES on the buckets should help me with the CSP work #414

Suggestions: 
`media.devportal.stage.mozit.cloud` for stage maybe
and 
`media.developer.mozilla.com` for prod

perhaps?

Your call ultimately, but hope those suggestions help

limed:
> Heya @limed - having CNAMES on the buckets should help me with the CSP work #414
> 
> Suggestions:
> `media.devportal.stage.mozit.cloud` for stage maybe
> and
> `media.developer.mozilla.com` for prod
> perhaps?
> 
> Your call ultimately, but hope those suggestions help

Yes we can do that, just know that cnaming a bucket requires me to rename the bucket (easier to recreate). I propose to name the media bucket as:

* `media.devportal.stage.mozit.cloud`
* `media.devportal.prod.mozit.cloud`

How about that?

stevejalim:
>    `media.devportal.prod.mozit.cloud`
> How about that?

As long as Mozilla policy is OK with that domain (mozit.cloud) being one that users will see requests too, I guess that's fine. Your call! :smile:

limed:
Hit a little snag, having a friendly name like `media.devportal.stage.mozit.cloud` that CNAMEs to the bucket will cause an SSL error[1]

>  Virtual host URLs are supported for non-SSL (HTTP) requests only. 

So unfortunately we will probably have to keep the `<bucketname>.s3.amazonaws.com`, that being said we can probably improve the situation by having a shorter bucket name. I propose the following:

* `devportal-media-stage.s3.amazonaws.com`
* `devportal-media-prod.s3.amazonaws.com`

[...]

PR #322, opened 2019-10-07
265: Support uploading user-media to AWS S3, instead of local filesystem

stevejalim:
This changeset enables the use _in production_ of `django-storages` to use AWS S3 for custom files/images uploaded into Wagtail, rather than store them in the local filesystem. (Local dev's default remains the same.)

Note that these user-media files go to a DIFFERENT S3 bucket from that configured for holding the baked-out static site.

When a site is built that features S3-hosted images, there is no need to download them only to upload them to the static-site bucket - the HTML and CSS just reference the images in the S3 user-media bucket instead. (It would be good to CNAME this user-media bucket at least - that's ticketed in #318)

This changeset also introduces a template for local settings that make it easier to develop S3-related behaviour while in local development mode (ie, reinstate some prod-like settings that `settings/dev.py` has disabled)

(Resolves #256)

## Key changes:

- add `django-storages` (BSD licence) to the project and configure
- ensure MEDIA_URL is `None` in production mode, so that `django-bakery` doesn't try to copy over files that aren't on the local filesystem
- fix generation of a base URL via `app_tags.get_scheme_and_host` to be aware of when user-media is in S3 (so its URLs are _already_ absolute), so that it doesn't produce a root URL prefix when it is not needed.
- add template `settings.local` with commented-out skeleton

## How to test

- check out local build
- copy `settings/local.py.example` to `settings/local.py`, uncomment all apart from the email backend one.
- fill in your own AWS credentials and ensure an S3 bucket exists for the user media (and, for consistency, also for the static site
- upload some images in Wagtail (http://localhost:8000/admin/images/) and confirm they end up in S3
- use those images in a page or two
- shell into a container (`docker-compose exec app /bin/sh`) and run `manage.py build` 
- inspect the output HTML in `/app/build/` and confirm that S3-hosted images are using a full S3 URI:

eg `<meta property="og:image" content="https://USER_MEDIA_BUCKET_NAME_HERE.s3.amazonaws.com/images/5.width-1200.png">` 
and
`      <div class="card-featured-image" style="background-image: url('https://USER_MEDIA_BUCKET_NAME_HERE.s3.amazonaws.co
m/images/5.width-1024.png')"></div>`

## Follow-up actions
- [ ] Re-upload the dozen or so images that are currently on the filesystem on staging, so that they work from S3 instead. (@stevejalim has downloaded them locally already)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ After installing Docker, use the **dev-setup** script to run the project locally
 
 This command will create an .env file with unique keys, build docker images and containers, run database migrations, and load fixture data.
 
+It's worth copying `settings/local.py.example` to `settings/local.py` but leaving everything commented out for now. Updating `local.py` can be a handy way to re-enable production-like settings (eg S3 uploads) that are disabled via the default local-development settings.
+
 ### Run locally
 
 With Docker Desktop running in the background, bring up the services by running:
--- a/developerportal/settings/base.py
+++ b/developerportal/settings/base.py
@@ -56,6 +56,7 @@ INSTALLED_APPS = [
     "wagtail.search",
     "wagtail.admin",
     "wagtail.core",
+    "storages",
     "bakery",
     "wagtailbakery",
     "modelcluster",
@@ -194,10 +195,6 @@ STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 STATIC_ROOT = os.path.join(BASE_DIR, "static")
 STATIC_URL = "/static/"
 
-MEDIA_ROOT = os.path.join(BASE_DIR, "media")
-MEDIA_URL = "/media/"
-
-
 # Django security settings (see `manage.py check --deploy`)
 
 CSRF_COOKIE_SECURE = True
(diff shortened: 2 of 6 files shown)


stevejalim:
(Note: I'll re-read this PR description tomorrow, to sense-check it)

liamnewmarch:
This PR is mostly AWS config which I’m not familiar with, so I’ll defer to the other reviewers!

stevejalim:
@limed Was just about to merge this, but realised I've only got one of the 11 pages of Images downloaded to my machine: https://developer-portal-stage.mdn.mozit.cloud/admin/images/

I can `kubectl cp` the images down to here, but wondering if there's a fast way to get them up to S3 that I have access to, or if it'll have to be either you doing a copy to S3 or me doing manual upload of 220 images....

stevejalim:
(We'll need to deal with this issue when moving to prod, too)

limed:
> @limed Was just about to merge this, but realised I've only got one of the 11 pages of Images downloaded to my machine: https://developer-portal-stage.mdn.mozit.cloud/admin/images/
> 
> I can `kubectl cp` the images down to here, but wondering if there's a fast way to get them up to S3 that I have access to, or if it'll have to be either you doing a copy to S3 or me doing manual upload of 220 images....

I've created a shell pod and moved the data from the EFS share to s3

PR #329, opened 2019-10-08
233: Attempt to make the k8s healthcheck run over HTTPS

### LINKED ISSUE #22: Create redirect from http -> https for developer-portal

### LINKED ISSUE #233: Address issue with readiness check when SSL redirect is enabled
https://github.com/mdn/developer-portal/pull/232 turned off the Django-level SSL redirect because it broke the k8s readiness/health checks.

Things to do:
- [ ] work out a way to make these checks work with SSL enabled
- [ ] re-enable SSL redirect at Django level

stevejalim:
Previously, when addressing #22  we hit an issue where the Kubernetes checks were failing. 

The traceback in that situation included a 301 happening, which _might be_ indicative that Django was trying to 301 a request from HTTP to HTTPS (it's unclear to me, from the single traceback available.)

So that the request avoids being 301ed from HTTP to HTTPS, which might be the root cause of health and readiness checks breaking.

Changes
- re-enabled Django-level redirect to HTTPS
- updated k8s deployment manifest to use HTTPS as the `scheme`, rather than the default of HTTP
- updated k8s deployment manifest to call URL paths with a trailing slash (as per their URLconf), just to help reduce the potential number of redirects involved
This changeset addresses ... by ...

(Resolves #233 )
(Resolves #22 )

## How to test

This one's going to be anothe fiddly one as we can only do it in staging. @limed what do you think about the approach here. Does it seem worth a shot, given the pain involved in reverting?


stevejalim:
@limed Given that you're the person best placed to see if it's working or not, would you mind merging and deploying it whenever suits you, please?

limed:
> On MDN, we have a hard shell (HTTPS) with a soft center (all HTTP), so we've never used HTTPS before for the liveness and readiness probes. This looks like the correct set-up to me though!

MDN doesn't have this problem because its fronted by CDN + ELB :)

stevejalim:
@limed Are you OK to take over the merge and release of this ticket, please?

PR #343, opened 2019-10-08
Don't push changes to eks cluster

limed:
We have 2 clusters at the moment, this change removes the need for the EKS cluster temporarily. The end goal is to use the EKS cluster at some point but we don't want to push changes there via CI atm.

## Key changes:

- Dont push CI changes to EKS cluster


limed:
r=limed

PR #344, opened 2019-10-08
Add k8s job to set media permissions

### LINKED ISSUE #340: Fix attachment permissions
Currently all attachments on the EFS/NFS mount is owned by root, it should be owned by the `devportaluser` user and group

### LINKED ISSUE #341: Mount attachments in celery worker
Worker needs attachments mounted on the pods.

limed:
This PR adds a k8s job that will create and chown the media directory. By default all folders get created with `root:root` permission, our containers run with GID and UID 1000 so this job fixes that. This job should only be needed once

(Resolves #340)

## Key changes:

- Add k8s job and Makefile targets to fix permission of media files

### DIFF
--- a/k8s/Makefile
+++ b/k8s/Makefile
@@ -82,6 +82,13 @@ k8s-rollback:
 k8s-history:
 	${KC} rollout history deploy ${APP_NAME}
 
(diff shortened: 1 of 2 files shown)


limed:
r=limed

PR #349, opened 2019-10-09
Implement redesigned tabbed panel component

dcgauld:
This changeset addresses issues around the tabbed panel component by implementing a cleaner, simpler design that scales better. Previously there were issues with margins, fixed heights and a generally cluttered look, whereas now information is easier to read and the panel can be used for a variety of things (seeing as only adding one item essentially displays the section as a banner).

There are also bug fixes and general improvements to the component throughout. *Note:* the design of the mobile version has remained unchanged.

Before:
Screen Shot 2019-10-09 at 6 09 39 PM

After:
Screen Shot 2019-10-09 at 6 09 28 PM

## Key changes

- Implement redesigned component and clean up related files.
- Remove unneeded tabbed panel heading field from topics.
- Fix spacing issues within pattern header and featured card containers.
- Fix a couple of small issues with the mobile design.

## How to test

- Add one tabbed panel item to a topic. Observe that the section looks correct.
- Continue to add more items, until the max of 3 and test responsively.
- Ensure both internal and external links work as expected.


src/js/organisms/tabbed-panels.js
```diff
@@ -41,36 +40,36 @@ module.exports = class TabbedPanels {
...
+
     if (!hash) return;
 
-    this.toggles.forEach(toggle => {
-      if (toggle.attributes.getNamedItem('href').value === hash) {
-        toggle.classList.add('highlight2');
-        toggle.classList.remove('highlight2-inverse');
+    const hashes = this.contentItems.map(el => `#${el.dataset.hash}`);
+
+    if (!hashes.includes(hash)) return;
```
liamnewmarch: I was going to ask if `includes` is supported in IE11 – it isn’t but we have a polyfill for it 👍

PR #360, opened 2019-10-10
Prevent 500 error when previewing a Video page associated with a Topic

### LINKED ISSUE #359: Previewing (Video?) pages seems to raise a 500
## Steps to reproduce

* Pick a page to edit in the Wagtail admin
* Preview it (with or without changing content)

## What I expected
* Seeing a rendered preview of the page

## What I got:
Screenshot 2019-10-10 at 14 51 55

liamnewmarch:
This change set addresses the 500 that occurs when trying to preview a Video with a Topic set.

The error occurs in the Video.related_resources method. Article has the same method with different logic, so I’ve copied that approach.

(Resolves #359 )

### DIFF
--- a/developerportal/apps/videos/models.py
+++ b/developerportal/apps/videos/models.py
@@ -263,7 +263,7 @@ class Video(BasePage):
     def related_resources(self):
         """Returns resources that are related to the current resource, i.e. live,
         public articles and videos which have the same topics."""
-        topic_pks = self.topics.values_list("topic")
+        topic_pks = [topic.topic.pk for topic in self.topics.all()]
         return get_combined_articles_and_videos(self, topics__topic__pk__in=topic_pks)
 
     def has_speaker(self, person):


developerportal/apps/videos/models.py
```diff
@@ -263,7 +263,7 @@ def read_time(self):
     def related_resources(self):
         """Returns resources that are related to the current resource, i.e. live,
         public articles and videos which have the same topics."""
-        topic_pks = self.topics.values_list("topic")
+        topic_pks = [topic.topic.pk for topic in self.topics.all()]
```
stevejalim: I can see how this caused the blow-up, in that the old version returned instances, not ids/pks.

Would be good to add/improve test coverage for all custom code, but as a fix for now, I trust @liamnewmarch that this is legit

liamnewmarch: > the old version returned instances, not ids/pks

Yes, it’s interesting that it worked at all.

PR #383, opened 2019-10-14
Add slack notifications for the jenkins build

### LINKED ISSUE #55: Jenkins job needs to notify on build success or failure
Add a notification that goes to slack and irc to notify us when build succeeds or fails

limed:
Adds slack notifications to developer-portal jenkins builds

(Resolves #55)

## Key changes:

- Adding slack notifications for the jenkins build pipeline

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,6 +41,11 @@ def deploy(config, environment) {
         make k8s-deployments
         make k8s-rollout-status
       """
+      notify_slack([
+        stage: "Deployed to ${environment}",
+        status: 'sucess'
+      ])
+
     } catch(err) {
       notify_slack([
         stage: "Failed to deploy to ${environment}",
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,7 +102,7 @@ node {
       break
 
     default:
-      stage ('Buld Image')  {
+      stage ('Build Image')  {
         sh "docker build --tag ${image}:${tag} ."
       }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -36,14 +61,26 @@ node {
   switch (env.BRANCH_NAME) {
     case 'master':
       stage ('Buld Image') {
```
stevejalim:
```suggestion
      stage ('Build Image') {
```

limed: Fixed :)

limed:
> I was thinking that you might want to add, within the `deploy` function, a call to `notify_slack` on success for a positive indication of success rather than just an absence of a failure messsage, but that depends on what you all want. Otherwise everything looks great!

I think thats a good point, I went ahead and added that

Jenkinsfile
```diff
@@ -16,13 +27,32 @@ def deploy(config) {
...
+      sh """
+        . config/${config}.sh
+        export APP_IMAGE_TAG=${tag}
+        make k8s-db-migration-job
+        make k8s-deployments
+        make k8s-rollout-status
+      """
+      notify_slack([
+        stage: "Deployed to ${environment}",
+        status: 'sucess'
```
escattone:
```suggestion
        status: 'success'
```

limed: Fixed in f839bab874e44aea0e3f4d82e3123cf588b5ec04

PR #398, opened 2019-10-16
Fixes certain race conditions on the ci test jobs

### LINKED ISSUE #389: CI job fails intermittently
During some builds the migrate task will fail because of the database not being up and ready yet and would fail with this error

```
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/base/base.py", line 217, in ensure_connection
    self.connect()
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/base/base.py", line 195, in connect
    self.connection = self.get_new_connection(conn_params)
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/postgresql/base.py", line 178, in get_new_connection
    connection = Database.connect(**conn_params)
  File "/usr/local/lib/python3.7/site-packages/psycopg2/__init__.py", line 126, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  the database system is starting up
```

@escattone suggested we use `urlwait` https://pypi.org/project/urlwait/ for this to wait and proceed only when the db becomes ready. Subsequent builds after this fail will also throw errors like the following

```
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/base/base.py", line 217, in ensure_connection
    self.connect()
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/base/base.py", line 195, in connect
    self.connection = self.get_new_connection(conn_params)
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/postgresql/base.py", line 178, in get_new_connection
    connection = Database.connect(**conn_params)
  File "/usr/local/lib/python3.7/site-packages/psycopg2/__init__.py", line 126, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: could not connect to server: Connection refused
	Is the server running on host "db" (172.27.0.4) and accepting
	TCP/IP connections on port 5432?
```

My suspicion here is that during the tear down process we never cleaned up the `.env` file properly which houses all the temporary passwords and also docker-compose is not tearing down the instance properly here. We can fix this my using `cleanWS()` in jenkins and running `docker-compose down --volumes --remove-orphans` before the build even happens

limed:
There are certain conditions that we run into that causes ci jobs fail
and its generally caused by either the database not being ready or a
permission denied error, which to me indicates that the old credentials
are still lingering after a job completes.

This commit hopes to fix that by using the `urlwait` tool to wait for
the database instance to be up and running as well as additional steps
in there that will go ahead and cleanup the environment before a build
and run.

(Resolves #389)

## Key changes:

- Add `urlwait` tool as a requirement
- Remove `.env` file if it exists
- Ensure that there are no running containers when `docker-compose` up gets run
- Add wait conditions in `ci-setup` script

## How to test

- From base directory run `scripts/ci-setup --build` and wait for build to complete
- Run `scripts/ci-tests`

Everything should work and there should be no errors when ran

### DIFF
--- a/scripts/ci-setup
+++ b/scripts/ci-setup
@@ -27,6 +27,12 @@ do
   shift
 done
 
+# Remove env file to get a clean env
+echo "Removing old .env file"
+if [ -f .env ]; then
+    echo "Removing env file"
+    rm -f .env
+fi
 
 echo "Setting up environment"
 
@@ -42,10 +48,18 @@ POSTGRES_PASSWORD=$(openssl rand -base64 64 | tr -d '+/\n=')
 POSTGRES_USER=admin
 EOF
 
+echo "Taking down any existing containers"
+docker-compose down --remove-orphans --volumes
+
 if [ -n "${ARG_BUILD}" ]; then
     echo "Running build"
     docker-compose up --build --detach
 fi
+
+# wait for services to be up
+echo "Waiting on services to come up"
+docker-compose exec -T app urlwait postgresql://db:5432/developerportal 30
+docker-compose exec -T app urlwait redis://redis:6379/0 30
 
 echo "Running migrate"
 docker-compose exec -T app python manage.py migrate
--- a/scripts/ci-setup
+++ b/scripts/ci-setup
@@ -29,7 +29,7 @@ done
 
 # Remove env file to get a clean env
 if [ -f .env ]; then
-    echo "Removing .env file"
+    echo "Removing old .env file"
     rm -f .env
 fi
 
--- a/scripts/ci-setup
+++ b/scripts/ci-setup
@@ -28,7 +28,6 @@ do
 done
 
 # Remove env file to get a clean env
-echo "Removing old .env file"
 if [ -f .env ]; then
     echo "Removing env file"
     rm -f .env
--- a/scripts/ci-setup
+++ b/scripts/ci-setup
@@ -29,7 +29,7 @@ done
 
 # Remove env file to get a clean env
 if [ -f .env ]; then
-    echo "Removing env file"
+    echo "Removing .env file"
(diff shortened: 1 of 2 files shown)


scripts/ci-setup
```diff
@@ -27,6 +27,12 @@ do
   shift
 done
 
+# Remove env file to get a clean env
+echo "Removing old .env file"
+if [ -f .env ]; then
+    echo "Removing env file"
+    rm -f .env
```
escattone: Good idea!

scripts/ci-setup
```diff
@@ -27,6 +27,12 @@ do
   shift
 done
 
+# Remove env file to get a clean env
+echo "Removing old .env file"
```
escattone: I would remove this one and just keep the one on line 33.

scripts/ci-setup
```diff
@@ -27,6 +27,12 @@ do
...
+if [ -f .env ]; then
+    echo "Removing env file"
```
escattone:
```suggestion
    echo "Removing old .env file"
```

scripts/ci-setup
limed: Opps good catch

scripts/ci-setup
limed: Fixed in c92cbc214415e8500ed1074bc3fae47f4fa661c0 and c3ee92a083a71cac25f27618bae233676dc32e26

scripts/ci-setup
limed: Fixed in a34864c9aed82291c1c5ce3f10786cab5d0c8389

PR #399, opened 2019-10-16
Fix some logical error that causes slack to notify even when tests pass

### LINKED ISSUE #55: Jenkins job needs to notify on build success or failure

limed:
Fixes a logic error that causes jenkins to notify even when tests passes

(Resolves #55)

## Key changes:

- Added a `catch` block
- Sneaked in a notify when tests are sucessful

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,12 +80,20 @@ node {
         try {
             sh "scripts/ci-setup --build"
             sh "scripts/ci-tests"
-        } finally {
-            sh "docker-compose down --volumes --remove-orphans"
+
             notify_slack([
+              stage: "Running test on tag ${tag}",
+              status: 'success'
+            ])
+
+        } catch (err) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -80,12 +80,20 @@ node {
...
-        } finally {
-            sh "docker-compose down --volumes --remove-orphans"
+
             notify_slack([
+              stage: "Running test on tag ${tag}",
+              status: 'success'
+            ])
+
+        } catch (err) {
+          notify_slack([
```
escattone: Add 2 spaces to indentation.
```suggestion
            notify_slack([
```

limed: Fixed in ac7272507f664aed0c3dcf0e5cc677dc41e8c609

PR #412, opened 2019-10-18
279: Re-architect how we're using our task queue to avoid concurrent builds

### LINKED ISSUE #279: Look at a way to purge outmoded "build-and-sync" tasks and just use the newest one.
I'll expand on this issue shortly, but adding it now while I write some docs about how the background build-and-sync-to-s3 behaviour will work.

stevejalim:
If merged, this changeset addresses the need to avoid concurrent and therefore conflicting builds being processed by the task queue.

It does this by switching from a pattern where builds are immediately enqueued (and picked up by workers) to one where builds are idempotently _requested_. A periodic task checks each minute if a build is needed and, if so, performs it.
 
(Resolves #279 - and goes beyond this, because it became clear it was necessary)

## Key changes:

-  Refactor how we enqueue jobs in Celery, to avoid concurrent builds and race conditions

    Prior to this commit, if a flurry of build requests were issued (eg multiple authors publishing)
this would put N jobs in the queue. Celery is configured by default to use the number of cores
on its worker node, so this would mean 4 (or more) builds taking place at once, and all targetting
the same output bucket in S3.

    IFF one could trust that the order of execution was stable, this would be tolerable (if wasteful)
but because it's async, there's no guarantee that Worker 4 will finish after Workers 1, 2 and 3.
Plus, this would cause four calls to invalidate the entire CDN, which itself is a bit of a black-box experience as it rolls from "InProgress" to "Complete".

    So, a new approach is needed.

    I've taken a few swings at this, most of which were just not satisfactory. (eg, in-task code that
checks for other running tasks of the same type is still prone to stale state and therefore bad
decisions about whether to terminate itself or carry on, etc). Running celery with `--pool=solo`
to only do one thing at a time would quickly cause problems as soon as we needed to do _other_
things with a task queue, or spin up multiple worker nodes, etc.

    So, instead, the approach has changed to one where builds are requested, and then that signal is checked frequently and exclusively actioned if it's neeeded.

    We do this using a sentinel value, stored in a shared (Redis-backed) cache that, when `True`,
signifies that a build is required. Separately from this, we have a regular scheduled task (via celery-beat) that calls the `_static_build_async` task every minute. That task function has been updated to:
a) check for the sentinel to decide whether to do a build or not, and - if it's set - to then un-set it before going ahead and start a build-and-sync to S3
b) use a distributed lock to ensure that no other builds can take place while this build is running
c) if a build is skipped because one _is_ running, ensure the sentinel is set again so that the next time the scheduler tries, it will still be clear that a fresh build is needed.

This means that for a flurry of, say, 8 requests, request 1 will trigger a build, while the rest will just (idempotently) set the sentinel to point out that someone else wanted a build while the current one was completing. If one or more fresh build requests come in while a build is in progress, that will be honoured as soon as the previous build completes.

- Update k8s config to spin up a celerybeat scheduler pod, too

- (implicit in the above) Start using Redis as a Django cache backend -- seemed appropriate given we already have it in place. Note we use DB `1`, not Celery's DB `0`

- Amend management command to call a more appropriate build-and-sync command, so that we don't risk concurrent builds

-  Add test settings based on production ones (not dev) and make the cache safe + ensure CI tests use the test settings

- Drop support for forced builds via this new approach (they are still possible in development, but they're not relevant to the production pattern anyway)

- Add tests

## How to test

- Ensure your .env has configuration for exporting a site to S3 in your `local.py` and that the bucket exists. *Do not use staging or production credentials/buckets here*.
```
# AWS_ACCESS_KEY_ID = "ADD_ME"
# AWS_SECRET_ACCESS_KEY = "ADD_ME"
# AWS_BUCKET_NAME = "BUCKET_NAME_HERE"  # Destination of baked site
# AWS_REGION = "REGION_NAME_HERE"  # eg eu-west-1
```
- Bring up the local docker stack: `docker-compose build && docker-compose up`
- Confirm you see `worker_1` and `scheduler_1` containers running
- Confirm you can see `scheduler_1` checking for the intent flag:

```
scheduler_1  | [2019-10-18 11:31:06,527: INFO/MainProcess] Scheduler: Sending due task add-every-minute (developerportal.apps.staticbuild.wagtail_hooks._static_build_async)
scheduler_1  | [2019-10-18 11:31:06,538: DEBUG/MainProcess] developerportal.apps.staticbuild.wagtail_hooks._static_build_async sent. id->921a5a92-8011-428b-a6b3-08c396ae11a6
scheduler_1  | [2019-10-18 11:31:06,541: DEBUG/MainProcess] beat: Waking up in 59.98 seconds.
worker_1     | [2019-10-18 11:31:06,542: INFO/MainProcess] Received task: developerportal.apps.staticbuild.wagtail_hooks._static_build_async[921a5a92-8011-428b-a6b3-08c396ae11a6]
worker_1     | [2019-10-18 11:31:06,549: INFO/ForkPoolWorker-1] [Static-build-and-sync] No fresh static build requested.
worker_1     | [2019-10-18 11:31:06,571: INFO/ForkPoolWorker-1] Task developerportal.apps.staticbuild.wagtail_hooks._static_build_async[921a5a92-8011-428b-a6b3-08c396ae11a6] succeeded in 0.02505229998496361s: None
```
- Open another terminal and shell into Django: `docker-compose exec app manage.py shell`
- `from developerportal.apps.staticbuild.wagtail_hooks import static_build`
- Now call `static_build()` to signal the intention for a static build, then wait to see the scheduler notice it and trigger a build-and-sync

```
worker_1     | [2019-10-18 11:33:23,953: INFO/MainProcess] Received task: developerportal.apps.staticbuild.wagtail_hooks._request_static_build[9485327f-de69-42fd-9a66-2e12759295dd]
worker_1     | [2019-10-18 11:33:23,954: INFO/ForkPoolWorker-1] [Static-build-and-sync requester] Caching a sentinel marker to request a static build within the next 60.0 seconds.
worker_1     | [2019-10-18 11:33:23,969: INFO/ForkPoolWorker-1] Task developerportal.apps.staticbuild.wagtail_hooks._request_static_build[9485327f-de69-42fd-9a66-2e12759295dd] succeeded in 0.014500599994789809s: None
scheduler_1  | [2019-10-18 11:33:24,533: INFO/MainProcess] Scheduler: Sending due task add-every-minute (developerportal.apps.staticbuild.wagtail_hooks._static_build_async)
scheduler_1  | [2019-10-18 11:33:24,536: DEBUG/MainProcess] developerportal.apps.staticbuild.wagtail_hooks._static_build_async sent. id->3dc85ce5-3436-4034-9977-c0d03a428e58
scheduler_1  | [2019-10-18 11:33:24,536: DEBUG/MainProcess] beat: Waking up in 59.99 seconds.
worker_1     | [2019-10-18 11:33:24,537: INFO/MainProcess] Received task: developerportal.apps.staticbuild.wagtail_hooks._static_build_async[3dc85ce5-3436-4034-9977-c0d03a428e58]
worker_1     | [2019-10-18 11:33:24,542: INFO/ForkPoolWorker-2] [Static-build-and-sync] Attempting fresh build
worker_1     | [2019-10-18 11:33:24,543: INFO/ForkPoolWorker-2] [Static-build-and-sync] Created temporary build dir /app/build/2019-10-18T11:33:24.543392+00:00
worker_1     | [2019-10-18 11:33:24,543: INFO/ForkPoolWorker-2] [Static-build-and-sync] (wagtail-bakery) command 'Build' started.
...
worker_1     | [2019-10-18 11:33:39,376: INFO/ForkPoolWorker-2] [Static-build-and-sync] (wagtail-bakery) command 'Publish' finished.
...
```

(Note that by default, the scheduler checks to see if a static build has been called once a minute, which makes it harder to see the locking in action. So, you can edit line 22 of `staticbuild/celery.py` to set the `"schedule": 3` - the build-and-sync takes more than 3 seconds, so we'll get to see the locking taking place)
- Now call `static_build()` to signal the intention for a static build, then wait to see the scheduler pick it up
- Immediately call `static_build()` again to signal the intention for a static build, then wait to see the scheduler come back a second or two later, spot the intention, but defer acting upon it because there's already a build in progress:

```
worker_1     | [2019-10-18 11:34:57,547: INFO/ForkPoolWorker-4] [Static-build-and-sync] Attempting fresh build
worker_1     | [2019-10-18 11:34:57,548: INFO/ForkPoolWorker-4] [Static-build-and-sync] Skipping this build attempt: another is in progress. Setting the sentinel so we'll try again later.
```

## MERGE BLOCKERS
- [x] need #422 sorting first, so that we can bring this up on k8s

### DIFF
--- a/developerportal/apps/staticbuild/wagtail_hooks.py
+++ b/developerportal/apps/staticbuild/wagtail_hooks.py
@@ -34,17 +34,40 @@ modeladmin_register(ArticleAdmin)
 
 # === TASK QUEUE / CELERY CODE ===
 
-# Concurrent builds, while wasteful, are possible when running Celery.
-# _request_static_build tries to deal with this.
+# A note on the approach here:
+#
+# We run the build-and-sync process async so that it doesn't block Wagtail
+# when a page is published. We also process the entire site each time, to avoid
+# the risk of some pages remaining stale but containing data from pages that
+# have since been updated (eg an Article that's also shown on a Topic page).
+# Full-replacement is the safest approach and doesn't need to be optimised yet.
+#
+# However, if we just added build tasks directly to the task queue, it'd be very
+# easy to end up with concurrent build-and-sync processes taking place, but with
+# noguarantee about the order in which they will complete, which risks the
+# "wrong" expected state being the one that ends up remaining published.
+# Also, it's worth avoiding unnecessary CPU and I/O costs for concurrent builds
+# that we don't need
+#
+# To this end, the approach we're taking is:
+#
+# * When a Page is published (which means we want a static build-and-sync), it
+#   spawns a Celery task which, when processed, simply sets a flag to say a
+#   "build is needed".
+#
+# * A separate periodic Celerybeat task checks for that flag and, if it finds
+#   it, takes down the flag and starts a static build
+#
+# * The static build runs within a simple distributed lock, so if a build it
+#   still running when the periodic task runs and finds a new build has been
+#   requestd, we don't allow a concurrent build, but we do preserve/re-instate
+#   the flag ready for the next attempt by the periodic task. This means we only
+#   ever have one build in progress, but we can immediately follow a build with
+#   a new one if there's still an unsatisfied request for a build.
 
 EXPECTED_BUILD_AND_SYNC_JOB_FUNC_NAME = (
     "developerportal.apps.staticbuild.wagtail_hooks._static_build_async"
 )
-
-# Evaluate this now to avoid an awkward edge cause during concurrent builds where it
-# (somehow) an os.path.join()ed value involving settings.BUILD_DIR gets re-set with
-# the joined value...
-BUILD_ROOT_DIR = settings.BUILD_DIR
 
 SENTINEL_KEY_TIMEOUT = (
     60 * 15
@@ -52,6 +75,11 @@ SENTINEL_KEY_TIMEOUT = (
 SENTINEL_KEY_NAME = "devportal-fresh-build-and-sync-needed"
 SENTINEL_LOCK_NAME = "deveportal-sentinel-lock"
 BUILD_LOCK_NAME = "deveportal-build-and-sync-lock"
+
+# Evaluate this now to avoid an awkward edge case during the former concurrent
+# builds where it (somehow) an os.path.join()ed value involving
+# settings.BUILD_DIR gets re-set with the joined value...
+BUILD_ROOT_DIR = settings.BUILD_DIR
 
 
 def _generate_build_path():
@@ -63,28 +91,27 @@ def _is_build_and_sync_job(job_details):
     return job_details["name"] == EXPECTED_BUILD_AND_SYNC_JOB_FUNC_NAME
 
 
-def _set_build_needed_sentinel():
+def _set_build_needed_sentinel(oid):
     """Idempotent flag that a static build should be attempted when next checked
     (which is within STATIC_BUILD_JOB_ATTEMPT_FREQUENCY seconds)"""
-    with cache.lock(SENTINEL_LOCK_NAME):
+    with redis_lock(SENTINEL_LOCK_NAME, oid):
         cache.set(SENTINEL_KEY_NAME, True, SENTINEL_KEY_TIMEOUT)
-        print(cache.get(SENTINEL_KEY_NAME))
 
 
-@app.task
-def _request_static_build(**kwargs):
+@app.task(bind=True)
+def _request_static_build(self, **kwargs):
     log_prefix = "[Static build requester]"
     logger.info(
         f"{log_prefix} Caching a sentinel marker to request a static build "
         f"within the next {STATIC_BUILD_JOB_ATTEMPT_FREQUENCY} seconds."
     )
-    _set_build_needed_sentinel()
+    _set_build_needed_sentinel(oid=self.app.oid)
 
 
 @app.task(bind=True)
 def _static_build_async(self, force=False, pipeline=settings.STATIC_BUILD_PIPELINE):
-    """Schedulable task that, if it gets the appropriate flag, calls each command
-    in the static build pipeline in turn.
+    """Schedulable task that, if it gets the appropriate flag as a sign to proceed,
+    calls each command in the static build pipeline in turn.
 
     See setup_periodic_tasks, above.
 
@@ -96,19 +123,17 @@ def _static_build_async(self, force=False, pipeline=settings.STATIC_BUILD_PIPELI
     build_dir = None
 
     if not force:
-        # Check to see if a build has been requested, and if so, wipe they key.
-        with cache.lock(SENTINEL_LOCK_NAME):
-            print("BEFORE cache.get(SENTINEL_KEY_NAME)", cache.get(SENTINEL_KEY_NAME))
+        # Check to see if a build has been requested, and if so, wipe the key.
+        with redis_lock(SENTINEL_LOCK_NAME, oid=self.app.oid):
             sentinel_val = cache.get(SENTINEL_KEY_NAME)
             cache.delete(SENTINEL_KEY_NAME)
-            print("AFTER cache.get(SENTINEL_KEY_NAME)", cache.get(SENTINEL_KEY_NAME))
             if sentinel_val is not True:
                 logger.info(f"{log_prefix} No static build requested.")
                 return
 
     # If we reach here, an actual build is needed, but only if one is not
     # already in progress.
-    logger.info(f"{log_prefix}Starting fresh build")
+    logger.info(f"{log_prefix} Starting fresh build")
     with redis_lock(lock_id=BUILD_LOCK_NAME, oid=self.app.oid) as lock_acquired:
         if lock_acquired:
 
@@ -142,7 +167,7 @@ def _static_build_async(self, force=False, pipeline=settings.STATIC_BUILD_PIPELI
                 f"{log_prefix} Skipping this build attempt: another is in progress. "
                 "Setting the sentinel so we'll try again later."
             )
-            _set_build_needed_sentinel()
+            _set_build_needed_sentinel(oid=self.app.oid)
 
 
 def static_build(**kwargs):
--- a/developerportal/apps/staticbuild/wagtail_hooks.py
+++ b/developerportal/apps/staticbuild/wagtail_hooks.py
@@ -100,7 +100,7 @@ def _set_build_needed_sentinel(oid):
 
 @app.task(bind=True)
 def _request_static_build(self, **kwargs):
-    log_prefix = "[Static build requester]"
+    log_prefix = "[Static-build-and-sync requester]"
     logger.info(
         f"{log_prefix} Caching a sentinel marker to request a static build "
         f"within the next {STATIC_BUILD_JOB_ATTEMPT_FREQUENCY} seconds."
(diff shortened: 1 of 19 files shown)


k8s/celery.yaml.j2
```diff
@@ -37,4 +37,24 @@ spec:
             limits:
               cpu: {{ CELERY_WORKER_CPU_LIMIT }}
               memory: {{ CELERY_WORKER_MEMORY_LIMIT }}
+        - name: {{ CELERY_SCHEDULER_NAME }}
```
stevejalim: @limed All this is a duplicate-and-tweak of the main `worker`, above in this file.

Can you please engage laser-eyes to check it makes sense (comparing with the change to `docker-compose.yml` will help) and if you're happy add the env vars, please (ticketed as #413)?

developerportal/settings/base.py
```diff
@@ -322,3 +322,16 @@
 CELERY_ACCEPT_CONTENT = ["application/json"]
 CELERY_TASK_SERIALIZER = "json"
 CELERY_RESULT_SERIALIZER = "json"
+CELERY_TIMEZONE = "UTC"
+CELERY_ENABLE_UTC = True
+
+_redis_cache_url = os.environ.get("REDIS_URL", "redis://redis:6379/0")
+_redis_cache_url = _redis_cache_url.replace("/0", "/1")
```
stevejalim: @limed let me know if this assumption of DB 0 will be a problem. Indeed, does the (secret) value for `REDIS_URL` in k8s end with a DB number? if not i will amend this code to handle that, too

limed: No the secret just uses the redis url and no `/0`

k8s/celery.yaml.j2
limed: This does look fine but what needs to happen as well is that we should set sane values in the `Makefile`

Issue #413 exposes it as a k8s env var but I think we should set this in the `Makefile`

See: https://github.com/mdn/developer-portal/blob/master/k8s/Makefile#L34-L40 we can set sane values here and then if need be override them under the `k8s/config/*.sh` files, does that make sense?

k8s/celery.yaml.j2
```diff
@@ -37,4 +37,24 @@ spec:
...
+          image: "{{ APP_IMAGE }}:{{ APP_IMAGE_TAG }}"
+          imagePullPolicy: {{ APP_IMAGE_PULL_POLICY }}
+          volumeMounts:
+            - name: developer-portal-fs
+              mountPath: {{ APP_MOUNT_PATH }}
```
limed: Do we still need this since we hosts our media under S3 now?

developerportal/settings/base.py
stevejalim: OK - will amend my PR to handle that. Thanks dude

k8s/celery.yaml.j2
stevejalim: Thanks. Will remove this code and we'll def go with your from PR #422

k8s/celery.yaml.j2
stevejalim: Celerybeat needs somewhere to use as a scheduler database of jobs. We could use django-celery-scheduler, which is a Django ORM backend for it, but then that means we [re]configure jobs manually, or via data migrations, so i went with the default scheduler, which is a simple on-disk instance of the python [shelve](https://docs.python.org/3/library/shelve.html) "database". So, if we're going to have more than one node/pod running celery-beat, it should be on a shared FS. If we're definitely going to cap it at one, then I think we'd be OK with a non-persistent disk

developerportal/apps/staticbuild/wagtail_hooks.py
```diff
@@ -27,41 +32,153 @@ class ArticleAdmin(ModelAdmin):
...
+    (which is within STATIC_BUILD_JOB_ATTEMPT_FREQUENCY seconds)"""
+    with redis_lock(SENTINEL_LOCK_NAME, oid=oid):
+        cache.set(SENTINEL_KEY_NAME, True, SENTINEL_KEY_TIMEOUT)
+
+
+def _get_build_needed_sentinel(oid):
+    with redis_lock(SENTINEL_LOCK_NAME, oid=oid):
+        sentinel_val = cache.get(SENTINEL_KEY_NAME)
+        cache.delete(SENTINEL_KEY_NAME)
+    return sentinel_val
```
escattone: @stevejalim Apologies again for being so slow to look at this! I like your overall approach! I've been skimming the details, and I had a question about the use here of `SENTINEL_LOCK_NAME`. It seems that lock is never checked anywhere to see whether or not it has been acquired? So it seems like the `SENTINEL_LOCK_NAME` lock is not really doing anything, or that it's not necessary, since it would never block the `cache.set(SENTINEL_KEY_NAME, ...` or `cache.get(SENTINEL_KEY_NAME)`. But I'm probably missing something!

stevejalim: Heya @escattone - thanks for the laser eyes!  

You're right - while in the main use of redis_lock we check for whether we actually get the lock, here (and in `_set_build_needed_sentinel`) I'm not. (It was a last-minute bit of clean-up refactoring and I clearly didn't think it through.)

The goal in `_get_build_needed_sentinel` was to try to make `cache.get()` and `cache.delete()` atomic in , given there's no `cache.pop()` operation available. (Looking back, it should probably be called `_pop_build_needed_sentinel` for greaster clarity, too).

Given that this is a situation where we really do want to store the sentinel, rather than only store if there's no existing lock, I guess we could  either:

a) just remove use of redis_lock and take the (pretty minor?) risk of another process setting the cache between `.get()` and `.delete()` (I say pretty minor cos the static site gets rebuild hourly anyway...)

b) change this to be a blocking/waiting lock, but that might deadlock (until the `redis_lock` times out)

So, I think opening a small PR (given this is now merged) to remove the call to `redis_lock` in this getter/setter pair (and the related `SENTINEL_LOCK_NAME`) is the way to go. Does that make sense to you? 

Thanks again for the catch!

escattone: @stevejalim Oh, I see, that totally makes sense that you'd want the `get` and `delete` calls within `_get_build_needed_sentinel` to be atomic. I think the [`MULTI`/`EXEC` transaction wrappers](https://redis.io/topics/transactions) that Redis provides may be exactly what you'd want here, but I don't think those commands can be made via the Django `cache` object. I think you'd have to use a dedicated Redis client like `redis-py`, which of course then you'd have to configure. However, you'd then have the full power of Redis available, including the `EVAL` command, which allows you to run entire scripts atomically (see https://redis.io/commands/eval#atomicity-of-scripts). I've used the `EVAL` command to handle locking in one of my projects(see https://github.com/escattone/walnut/blob/master/walnut/core.py if you'd like to see examples).

escattone: @stevejalim If setting up a dedicated Redis client seems like more work than you think it's worth, your idea of removing the call to `redis_lock` in the getter/setter pair seems good to me as well. In other words, it's so tempting to be a purist, but it just may not be worth it! :)

stevejalim: Yeap, we're def on the same page @escattone! Would be nice to do it 'right', but maybe it's not 100% needed... FWIW, i originally was using django-redis's `cache.lock()` which maps to `redis-py`'s own `Lock`, but thought it better to standardise on one lock implementation (and then fumbled the refactor :/)

FWIW here's the django-redis lock: https://niwinz.github.io/django-redis/latest/#_locks which might actually be a reasonable drop-in, now that the project is using `django-redis` anyway. `.lock` is basically a pass-through to `redis.py`

No need to reply - i know you're busy. I've made a new issue to come back v soon (P2, moving soon to P1) and address the point you've raised. Thanks for the support!

escattone: @stevejalim Wow, didn't know about django-redis' `cache.lock()` and `redis-py`'s `Lock`! Really nice! Thank you!

PR #415, opened 2019-10-18
Rename media bucket

### LINKED ISSUE #318: Add CNAME (and/or CDN) for user-media in S3

limed:
Simplifies the name of the s3 bucket, currently its using the convention of `developer-portal-<environment>-<account id>`, this PR moves the media bucket to a different bucket that has a shorter name

(Partially Resolves #318)

## Key changes:

- Rename media bucket


limed:
r=limed

PR #434, opened 2019-10-22
Support HTTPS for local development.

### LINKED ISSUE #424: Support HTTPS in local development
It would be useful to have HTTPS parity with real deployments while doing local builds - it'd make things like CSP easier to check/extend etc, too

stevejalim:
This changeset addresses a need for serving the local build using HTTPS.

It adds nginx + SSL cert generation into the local docker stack, following [how we do it for kuma](https://github.com/mdn/kuma/blob/master/docker-compose.ssl.yml)

Note that SSL is now the default for the local stack, but http://localhost:8000 still responds.  
Ideally all local development should be done on https://developer-portal-127-0-0-1.nip.io so that future changes (eg CSP support) don't break belocal behaviour. If you want to work offline, you can add `127.0.0.1    developer-portal-127-0-0-1.nip.io` to your `/etc/hosts`

Local fixtures are updated so that the default Wagtail site is expecting to be on `developer-portal-127-0-0-1.nip.io`, and this is matched by a new default for `BASE_URL`, so that the static build still works.

Tried a static build and it still works fine.

Nothing in this changeset should affect non-local behaviour.

(Resolves #424)

## How to test

- `docker-compose down` of your local build, if running
- check out branch
- `docker-compose build && docker-compose up`
- visit the site at `https://developer-portal-127-0-0-1.nip.io`
- visit the CMS at `https://developer-portal-127-0-0-1.nip.io/admin/` - you'll need to re-authenticate because the cookie domain is different
- visit both at `http://localhost:8000` to see you can still get to it the old way (Wagtail should work exactly the same, but the rendered site may break once we add CSP rules etc)


stevejalim:
@dcgauld Not expecting you to review this at the mo, but pinging you so that you're aware we'll have a HTTPS local stack when you're back on the project. 👋

stevejalim:
@escattone If you have any experience with a similar issue of CI failing in kuma when nginx+HTTPS is involved, I'd welcome hearing about it.

@limed Given the command for booting up `app` involves generating the SSL certs, maybe that's what's breaking `app_1` ?

stevejalim:
TBH, if this is proving a pain, we can live without it, but it would make other dev easier, ultimately

limed:
> @limed Given the command for booting up `app` involves generating the SSL certs, maybe that's what's breaking `app_1` ?

Resolved this by taking the whole cert generation out of `app` we now generate the cert using a tool called `cfssl` and it runs once each time a `docker-compose up` is invoked. Commit 94893abfbe37db80c1320fbec19ca4674fe50cc6 resolves this

stevejalim:
@limed - thanks so much for the work on this! I really apprecaite it.

Checked it out locally and it works fine, which is great.

I can't approve it because technically I created it, so if you could and merge it, that would be :star:

etc/nginx/ssl/README.md
```diff
@@ -0,0 +1,7 @@
+# SSL support for the local build
+
+This folder will contain self-signed certificates for SSL support in local development.
+
+They are generated/updated each time the local Docker stack is upped using `etc/nginx/ssl/gen-cert.sh`
```
stevejalim: @limed Can you update this line of the README please? It looks like it refers to the old/broken way I initially tried

etc/nginx/ssl/cfssl.json
```diff
@@ -0,0 +1,20 @@
+{
+  "CN": "developer-portal-127-0-0-1.nip.io",
+  "hosts": [
+    "developer-portal-127-0-0-1.nip.io",
+    "demos.developer-portal-127-0-0-1.nip.io"
```
stevejalim: @limed Any particular reason to havee this extra subdomain?

limed: No particular reason, I copied the domains from the openssl config file we had. If its not needed it can be removed

etc/nginx/ssl/README.md
limed: Actually this is the update method, the `gen-cert.sh` script is similar to the old way but the content of the script is different

.travis.yml
```diff
@@ -21,7 +21,7 @@ before_install:
   - docker-compose version
 
 install:
-  - scripts/ci-setup --build
+  - scripts/ci-setup --build --debug
```
limed: Remove debug line since we are done debugging

limed: Removed in 3b21e11b8ab2129e8ef41220ffc0b83b45c87d46

etc/nginx/ssl/cfssl.json
limed: Removed in 3b21e11b8ab2129e8ef41220ffc0b83b45c87d46

PR #445, opened 2019-10-23
Remove EFS references, we no longer serve media out of EFS

### LINKED ISSUE #444: Remove shared EFS volume
Since moving to S3 backed media storage we no longer need the EFS share anymore. This should be removed from terraform as well as K8S

limed:
Removes dependence on EFS, removing the Persistent Volume and Persistent Volume claim from the codebase.

(Resolves #444 )

## Key changes:

- Removes shared volume from k8s


limed:
r=limed

Issue #369, opened 2019-10-11
Get ahead of heavily stuffed "directories" in S3


stevejalim:
At the moment, the `mozimage.models.MozImage` model uses the default `upload_to` path for the image. This works, of course, but there's an issue we might need to get ahead of, if we will ever need to browse those directories: all the files go into the same virtual directory / namespace, so listing/browsing the contents of a bucket can be overwhelming and might result in a client timeout in a worst case.

Is this something we need to think about avoiding? Will we be browsing the S3 bucket directly?

(One way to avoid it is to add a custom callable that generates a hash as a prefix for the file, then uses the first n chars of that hash as a prefix/shard and add it to the directory path.

eg: `abc/abcdef123456/foo.jpg`, `ddd/ddddef223344/bar.jpg`

(Though this change to the `upload_to` path would require moving existing files in S3, but that's not the end of the world)

limed:
> @limed @escattone No rush, but what do you guys think of this? A problem or a non-problem? How do other projects at Mozilla organise their S3 assets?

I don't know how other projects in mozilla does this but I like the idea of using a hash as a prefix to the file.

To answer your question about whether we will be browsing the s3 bucket directly I don't think we will do it a lot unless we are trying to debug an issue, so if this takes lots of dev cycles I wouldn't worry about it. But I do like the idea :)

stevejalim:
Thinking about this more, it would be handy to do "just in case", and would be similar to how things are done in MDN (there they use `YY/MM/DD` 'path sharding' which is an easier approach to browse through).

Doing this before we hit proper production would also make life a bit easier. Even if prod starts as a clone of the staging DB and bucket, there won't be that many files to move.

Will need a three-phase/three-releases approach: 

* Release 1: For each image field, we add a new callable/method for determining the `upload_to` path for it, _but this is not wired in to the model yet_. 
   A one-off script/management command will iterate through all images and copy the image in S3 from the current path to wherever the relevant new callable says it should be, based on the new logic. 
* Release 2: This will contain the changes that enable the custom `upload_to` logic in the relevant models (including a no-schema-change Django Migration for each). Once live, Wagtail will automatically look down the new paths for the files (which were copied over in the prev deploy) so it'll be a seamless swap-over.
* Release 3: Contains a management command that deletes the original, unmoved images from the root path of the S3 assets bucket -- and we should run this once we're happy that the previous two releases have no flaws.

[...]

PR #446, opened 2019-10-23
Support filename hashing for S3 uploads

stevejalim:
This changeset adds a custom `get_upload_to` method on `mozimages'` `MozImage` that takes the original filename and instead uses a slightly truncated SHA1 hash of it instead. This hashed filename is then respected in Renditions of the original image, too.

This change does NOT break existing files hosted on S3 but it may well break any image links on a local dev build that is not using S3.

Note that settings.AWS_S3_FILE_OVERWRITE is set to True to overcome django-storages default - see https://github.com/wagtail/wagtail/issues/1686

(Partiallt deals with #369)

Tested locally against my own S3 bucket.

### DIFF
--- a/developerportal/apps/mozimages/models.py
+++ b/developerportal/apps/mozimages/models.py
@@ -1,3 +1,5 @@
+import hashlib
+
 from django.db import models
 
 from wagtail.images.models import AbstractImage, AbstractRendition, Image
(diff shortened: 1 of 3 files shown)


callahad:
Discussed this with Steve on Slack, and I think we need a decision from product:

> Do we need to obscure original file names at all?

If yes, then any kind of hash / uuid / renaming scheme is more or less equally valid.

If no, then we can skip this PR.

Either way we should separately include the `AWS_S3_FILE_OVERWRITE` fix.

PR #454, opened 2019-10-25
Deploy to 2 clusters

### LINKED ISSUE #453: Attempt to deploy to multiple clusters
We want to at some point migrate to EKS, our current deployments only deploys to a singular cluster and in order to make our lives easier we should deploy to both clusters in the interim.

limed:
This PR deploys developer-portal to 2 clusters (our existing KOPS cluster and an EKS cluster), however the only difference is that one of the cluster will not have any `celery-beat` and `celery-worker` pods running.

(Resolves #453)

## Key changes:

- Create config for a second cluster
- Update `Jenkinsfile` to deploy to another cluster

### DIFF
--- a/k8s/config/prod.sh
+++ b/k8s/config/prod.sh
@@ -1,17 +1,17 @@
 #!/usr/bin/env bash
-echo '--> Setting environment to developer-portal STAGE on mdn-apps-a cluster'
+echo '--> Setting environment to developer-portal PROD on mdn-apps-a cluster'
 
 export KUBECONFIG=${HOME}/.kube/mdn-apps-a.config
 
 # Define defaults for environment variables that personalize the commands.
-export TARGET_ENVIRONMENT=stage
+export TARGET_ENVIRONMENT=prod
 export K8S_NAMESPACE=dev-portal-${TARGET_ENVIRONMENT}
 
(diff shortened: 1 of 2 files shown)


k8s/config/prod.sh
```diff
@@ -0,0 +1,31 @@
...
+export APP_MEMORY_REQUEST=2Gi
+export APP_GUNICORN_WORKERS=2
+export APP_HOST=developer-portal.prod.mdn.mozit.cloud
+export APP_EXPORTED_SITE_HOST=developer-portal-published.prod.mdn.mozit.cloud
+export APP_AWS_BUCKET_NAME=developer-portal-prod-178589013767
+export APP_AWS_STORAGE_BUCKET_NAME=devportal-media-prod
+export APP_AWS_BUCKET_REGION=us-west-2
+
+export CELERY_SCHEDULER_REPLICAS=0
+export CELERY_WORKER_REPLICAS=0
```
escattone: Oh wait, just noticed these were both zero. Did you want these to be something non-zero for prod?

limed: Yes its on purpose, the older cluster has the `celery-worker` and `celery-beat` running and in order for us to not go running around to figure out which cluster the task ran from we set one of the cluster to `0` so that only one cluster is running the tasks

Issue #475, opened 2019-10-28
Disable table-dropping migration once SSO rollout is complete  


stevejalim:
#410 contains a schema migration (`common/migrations/0001_drop_social_auth_app_tables.py`) that removes the social-auth tables that its changeset makes redundant.

Once we're sure that the changeset has been deployed to stage and prod, we should disable that migration so that if, in the future, we re-add social-auth to the project, test runs and clean installs won't "mysteriously" end up with missing DB tables because of that migration

Issue #469, opened 2019-10-25
Make use of Wagtail 2.7's new flag to stop users changing their email


stevejalim:
As of 2.7 (scheduled for 6 November), Wagtail will add `WAGTAIL_EMAIL_MANAGEMENT_ENABLED` which stops users changing their email address. This is a good thing to add, because it stops them breaking SSO, amongst other things.

We should enable this new setting as soon as it lands

AC:
- [ ] we have a process for changing your email address

PR #493, opened 2019-10-28
401 Add support for Mozilla OpenID Connect SSO (via Auth0)

### LINKED ISSUE #401: Support LDAP SSO for Wagtail and Django admins
Confirmation and spec to come, but this may well be essential pre-alpha

stevejalim:
This changeset addresses the need for SSO as the _only_ way in to the CMS or the Django Admin. It does this by adding [mozilla-django-oidc](https://mozilla-django-oidc.readthedocs.io/) as an authentication backend and also removing support for conventional Django model-backed auth.

Once merged, this changeset means the default way to authenticate with the admins is via SSO. (A workaround is possible).

User-creation-via-SSO is deliberately disabled. This changeset also means that a User account with an email that corresponds to the human user's Mozilla SSO identity MUST be present in Wagtail for them to be signed in.

Note that this PR touches a few areas, so is probably best reviewed commit-by-commit. I've also included easy-to-follow test steps for checking locally.

When reviewing, please wear your most sceptical hat.

(Resolves #401 )

## Key changes:

- Remove existing Github social-auth login/account creation + drop relevant database tables
  - social-auth was added in early development, but hasn't been used and isn't the best fit for what we now know  we need. The app has been removed from config and dependencies and a schema migration has been added to drop those tables from staging and production as a precaution.
  - There's a ticket (#475) to disable this migration once it's no longer useful

- Add [mozilla-django-oidc](https://mozilla-django-oidc.readthedocs.io/) to the project, including:
  - adding a new build dep to the Dockerfile
  - configuring additional settings, expecting to find them in local.py or env vars
  - Plugging in appropriate logging as per `mozilla-django-oidc` docs

- Plug in of Auth0 to allow SSO for Wagtail Admin and Django Admin

- Remove support for using Django's `ModelBackend` for auth (and document how to add it back in for local development if you lack a Mozilla SSO account)
  - Fix/address test regressions caused by dropping the `ModelBackend`

- Remove the use of a password field when managing Wagtail Users. (This means new users get an unusable, blank password by default and cannot change it to anything usable). It's SSO or no-entry, basically. Locking down email changes is also ticketed, once it becomes available (#469)

- Override Wagtail and Django-Admin auth HTML templates to not present username and password fields because they are now redundant and confusing 

- Add tests to prove login forms are gone AND that POSTing to the URLs won't work anyway

- Update documentation

## Screenshots

Screenshot 2019-10-28 at 11 32 21

Screenshot 2019-10-28 at 11 32 45

### Setup

* Start on `master`:
    - [ ] If needed: `docker-compose exec app python manage.py createsuperuser` - and create a user whose email address is the same as your Mozilla SSO identity. 
        - OR edit the existing superuser in Wagtail (`/admin/users/`) to use that email address.
    - [ ] Sign in to Wagtail (at `/admin/`) with that admin identity using the username and password you specified.
    - [ ] Create a SECOND user (at `/admin/users/`), using a different email address.
        - [ ] Give them an Editor role, **not** an Admin one

Screenshot 2019-10-28 at 12 17 50

- [ ] Sign out

* Switch to this PR's branch:
    - [ ] Run migrations, which will drop the social-auth tables from your local DB (no longer needed, but good to have someone else run the migration)
    - [ ] Add/update your `settings/local.py` with the OIDC-related configuration from the lower part of `settings/local.py.example` (I will send these to reviewers securely)

### Test SSO
- [ ] Go to `/admin/` and check you can sign in to Wagtail via your SSO account
- [ ] Sign out
- [ ] Go to `/django-admin/` and confirm you can sign in to Wagtail via your SSO account. (Note that it redirects you to the Wagtail admin, but if you update the URL to `/django-admin/`, you will be authenticated and see the Django Admin. This minor niggle is tolerable, given how rarely most users will need the Django Admin.)
- [ ] Sign out

### Test when there is no user in Wagtail that matches the identity provided by SSO
(This shows that drive-by signups/automatic user creation has been disabled)

- [ ] Open a Django shell with `docker-compose exec app python manage.py shell` and change the email associated with the first user to no longer be the one that maps to you Mozilla SSO identity:
```
>>> from django.contrib.auth.models import User
>>> one = User.objects.first()
>>> one.email = "test2@example.com"
>>> one.save()
```
- [ ] Now try to log in to `/admin/` with SSO - you should be sent to `/` with no messaging. While this experience isn't great, it's somewhat out of our hands and fine for alpha, given all contributors should have been onboarded first.

- [ ] Try to log in to `/django-admin/` with SSO - you should also be sent to `/` with no messaging.

- [ ] In the Django Shell, confirm there are still only two Users

### Test can't access Django Admin if not an Admin

- [ ] Sign out of the CMS if you are signed in
- [ ] Re-open / re-use your Django shell and give the SECOND user (the one who is just an Editor) the email address associated with your Mozilla SSO identity:
```
>>> from django.contrib.auth.models import User
>>> one = User.objects.first()
>>> two = User.objects.last()
>>> one.email = "test@example.com"
>>> one.save()
>>> two.email = "<YOUR_LOCAL_HANDLE_HERE>@mozilla.com"
>>> two.save()
>>> one.email
'test@example.com'
>>> two.email
'<YOUR_LOCAL_HANDLE_HERE>@mozilla.com'
```
While you're here, also confirm that this user has no staff or superuser flag set:
```
>>> two.is_staff
False
>>> two.is_superuser
False
```
- [ ] Confirm you **can** sign in to `/admin/` (ie, Wagtail) using SSO
    - [ ] Confirm this is for the **Editor** identity, not the original admin
    - [ ] Confirm you don't see a 'Settings' option in the left-hand menu

Screenshot 2019-10-28 at 12 18 17

- [ ] Sign out
- [ ] Try signing in to `/django-admin/` using SSO
    - [ ] Confirm you get authenticated and redirected to Wagtail's `/admin/`, but if you go back to `/django-admin/`

### [Stretch] Test can log in in dev mode

If you have time, you can try re-enabling 'traditional' login via username and password, which is now supported only via explicit, local configuration (ie, the default is SSO only and it requires the deliberate setup of config in `settings/local.py`, so cannot happen in production)


stevejalim:
@escattone I appreciate you're plate-spinning right now, but when you have time for devportal, I'd really appreciate it if you could prioritise this one ahead of any other PRs, because I'd like to ship this by the middle of this week, if possible. I'll email you the relevant credentials (PGPed) for a local test-drive, too. Cheers, sir!

Dockerfile
```diff
@@ -24,6 +24,7 @@ RUN apk add --no-cache --virtual .build-deps \
 RUN apk add --no-cache \
   jpeg-dev \
   libc-dev \
+  libffi-dev \
```
stevejalim: @limed This addition to the Dockerfile is needed to allow `mozilla-django-oidc` to be installed. OK with you from an Ops point of view?

developerportal/apps/articles/tests/test_articles.py
```diff
@@ -1,10 +1,10 @@
-from wagtail.tests.utils import WagtailPageTests
+from developerportal.apps.common.test_helpers import PatchedWagtailPageTests
```
stevejalim: There are quite a few places where this simple replacement of the base test case happens - it's just because `WagtailPageTests` have a `login()` step that expects Django's `ModelBackend` to be available, but we've dropped it entirely. 

As such `PatchedWagtailPageTests` gets around this missing backend issue in order to make the Wagtail test-helpers in `WagtailPageTests` viable again. You'll see the code change later

developerportal/apps/common/migrations/0001_drop_social_auth_app_tables.py
```diff
@@ -0,0 +1,38 @@
...
+        "social_auth_code",
+        "social_auth_nonce",
+    ]
+    logger.info("Dropping redundant DB tables %s", table_names)
+    with connection.cursor() as cursor:
+        for table_name in table_names:
+            query = f"DROP TABLE IF EXISTS {table_name}"
+            logger.info(f"Query: {query}")
+            cursor.execute(query)
+
```
stevejalim: This works when the tables exist and also when they have already gone (or never existed in the first place - eg, in CI).

An example output from tests running:
```
Creating test database for alias 'default'...
INFO:developerportal.apps.common.migrations.0001_drop_social_auth_app_tables:Dropping redundant DB tables ['social_auth_usersocialauth', 'social_auth_partial', 'social_auth_association', 'social_auth_code', 'social_auth_nonce']
INFO:developerportal.apps.common.migrations.0001_drop_social_auth_app_tables:Query: DROP TABLE IF EXISTS social_auth_usersocialauth
INFO:developerportal.apps.common.migrations.0001_drop_social_auth_app_tables:Query: DROP TABLE IF EXISTS social_auth_partial
INFO:developerportal.apps.common.migrations.0001_drop_social_auth_app_tables:Query: DROP TABLE IF EXISTS social_auth_association
INFO:developerportal.apps.common.migrations.0001_drop_social_auth_app_tables:Query: DROP TABLE IF EXISTS social_auth_code
INFO:developerportal.apps.common.migrations.0001_drop_social_auth_app_tables:Query: DROP TABLE IF EXISTS social_auth_nonce
System check identified no issues (0 silenced).
```

stevejalim: I've ticketed-up removing this migration once it's done its duty: https://github.com/mdn/developer-portal/issues/475

developerportal/apps/common/test_helpers.py
```diff
@@ -0,0 +1,17 @@
...
+
+class PatchedWagtailPageTests(WagtailPageTests):
+    @override_settings(
+        AUTHENTICATION_BACKENDS=(
+            "mozilla_django_oidc.auth.OIDCAuthenticationBackend",
+            "django.contrib.auth.backends.ModelBackend",
+        )
+    )
+    def login(self):
+        return super().login()
```
stevejalim: This approach - to patch back in the `ModelBackend` temporarily - is the least painful/best-fit way I could come up with, given that there's no mock/dummy included in the OIDC authentication backend and I didn't have time to write one. 

Reviewers: let me know if it makes your eye twitch

developerportal/pipeline.py
```diff
@@ -1,65 +0,0 @@
-import os
```
stevejalim: This file became irrelevant when we dropped the Github/social auth

developerportal/settings/base.py
```diff
@@ -336,3 +314,55 @@
...
+# Mozilla OpenID Connect / Auth0 configuration
+
+# How frequently do we check with the provider that the
+# user still exists and is authorised?
+OIDC_RP_SIGN_ALGO = "RS256"
+
+OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS = (
+    15 * 60
+)  # 15 mins is the default, but re-expressed here for visibility
+OIDC_CREATE_USER = False  # We don't want stop drive-by signups
```
stevejalim: Oops. Typo fix!
```suggestion
OIDC_CREATE_USER = False  # We don't want drive-by signups
```

developerportal/templates/admin/login.html
```diff
@@ -0,0 +1,42 @@
...
+{% block content %}
+
+  {% use_conventional_auth as conventional_auth_enabled %}
+
+  {% if conventional_auth_enabled %}
+
+    {{block.super}}
+
+  {% else %}
+
```
stevejalim: I wanted to keep the error messages visible, just in case, but the Django Admin base template didn't really have a great block-inheritance pattern to lean on.

developerportal/urls.py
```diff
@@ -16,7 +16,7 @@
     url(r"^admin/", include(wagtailadmin_urls)),
     url(r"^documents/", include(wagtaildocs_urls)),
     url(r"^article-feed/", RssFeeds()),
-    url(r"^oauth/", include("social_django.urls", namespace="social")),
+    url(r"^auth/", include("mozilla_django_oidc.urls")),
```
stevejalim: Perhaps not _the best_ path naming convention, but at least it's not clashing with Django or Wagtail's own urls

Dockerfile
limed: This sounds good from my point of view :+1:

developerportal/settings/base.py
```diff
@@ -336,3 +314,55 @@
...
+
+# Mozilla OpenID Connect / Auth0 configuration
+
+# How frequently do we check with the provider that the
+# user still exists and is authorised?
+OIDC_RP_SIGN_ALGO = "RS256"
+
+OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS = (
+    15 * 60
+)  # 15 mins is the default, but re-expressed here for visibility
```
callahad: Do we need to have active code here at all, or can we just leave a comment along the lines of:

```
# Tokens expire every 15 minutes by default
# To override, set OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS
```

developerportal/apps/common/test_helpers.py
callahad: It's gnarly, but it's straightforward and contained.

Add a comment around `developerportal/settings/base.py#L120` noting that backends are overridden in tests to help with potential future changes to `AUTHENTICATION_BACKENDS`?

developerportal/settings/base.py
```diff
@@ -336,3 +314,55 @@
...
+OIDC_CREATE_USER = False  # We don't want stop drive-by signups
+
+OIDC_RP_CLIENT_ID = os.environ.get("OIDC_RP_CLIENT_ID")
```
stevejalim: TODO: amend these so that only the secret and ID come from the env -- @limed has confirmed that @jdow from the IAM team is fine with the full OIDC URLs living in the codebase. (This makes sense, too, because they are not secret, given that anyone will know them after clicking the SSO sign in button)

developerportal/templates/admin/login.html
```diff
@@ -0,0 +1,42 @@
...
+    {% endfor %}
+    {% endif %}
+
+    <p class="module">
+      <a class="button" href="{% url 'oidc_authentication_init' %}">
+        {% trans 'Sign in with Mozilla SSO' %}
+      </a>
+    </p>
+    <p>
+      Note that after sign-in, you will be sent back to the CMS admin. Please re-access {% url 'admin:index' %} manually.
```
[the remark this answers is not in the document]
stevejalim: Not a super- clean way, because Django only takes a string (a path or URL name) for `LOGIN_REDIRECT_URL`. I _could_ add a custom view to handle the redirect and send people the right way, but given the very low likelihood of authorised people wanting the Django Admin itself (rather than just Wagtail's Admin), I thought a bit of help text + human brainpower will bridge the gap for now. We can revisit this post-alpha if it annoys people.

developerportal/settings/base.py
```diff
@@ -336,3 +314,55 @@
         "OPTIONS": {"CLIENT_CLASS": "django_redis.client.DefaultClient"},
     }
 }
+
+# Mozilla OpenID Connect / Auth0 configuration
+
+# How frequently do we check with the provider that the
+# user still exists and is authorised?
```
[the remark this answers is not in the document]
stevejalim: Great call - thanks!

PR #522, opened 2019-10-28
Update cert arn to point to new cert

limed:
Updating certificate ARN for ELB and also adding additional Tags for ELB


limed:
r=limed

PR #554, opened 2019-10-29
547: Update navigation

### LINKED ISSUE #547: Update navigation
* The Docs navigation item currently links out to MDN Web Docs. Change the wording of the link to MDN Web Docs. Possibly include an icon like we do for Firefox. 

* The Send Feedback link was designed for the internal alpha & should be removed from the navigation bar.
  * If feedback is still desired from external users, we could love this link to the footer.

stevejalim:
This changeset addresses issues with the nav by:

* Removing "Send Feedback" link from the top nav
* Rephrasing Docs as MDN Web Docs and add the Mozilla "M" logo beside it
(Resolves #547)

## Before
Screenshot 2019-10-29 at 21 37 18

## After
Screenshot 2019-10-29 at 21 42 04

The only question that needs answering is: is the "M" logo on-brand for MDN, or does it need to be the dino logo? (If so, we'll need that sourcing and adding to the developer portal repo)


stevejalim:
@atopal Feel free to bounce this to Ali if you think she should sign off

PR #596, opened 2019-10-30
591: only load GA code if Google Analytics ID is available

### LINKED ISSUE #591: Support no GA code being configured
As prep for #101 we should make sure that we gracefully handle the absence of a GA code

stevejalim:
This changeset prepares the ground for switching the Google Analytics ID from staging to production, so there will (normally) be no GA ID configured. In that situation, we want to avoid including any GA code, rather than risk it complaining in the console. 

(Resolves #591)

## Key changes:

- Only load GA code if Google Analytics ID is available

## No GOOGLE_ANALYTICS code configured in environment

```
...
  </footer>

    <script>window.Mzp = {}</script>
    <script src="/static/js/bundle.js"></script>

  </body>
</html>
```

## With a GOOGLE_ANALYTICS code configured in environment

```
...
  </footer>

    <script>window.Mzp = {}</script>
    <script src="/static/js/bundle.js"></script>

    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-[REDACTED]-1"></script>
    <script>
      window.dataLayer = window.dataLayer || [];
      function gtag(){dataLayer.push(arguments);}
      gtag('js', new Date());
      gtag('config', 'UA-[REDACTED]-1');
    </script>

  </body>
</html>
```

### DIFF
--- a/developerportal/templates/base.html
+++ b/developerportal/templates/base.html
@@ -62,6 +62,7 @@
     {% endblock %}
     <script>window.Mzp = {}</script>
     <script src="{% static 'js/bundle.js' %}"></script>
+    {% if GOOGLE_ANALYTICS %}
     <script async src="https://www.googletagmanager.com/gtag/js?id={{ GOOGLE_ANALYTICS }}"></script>
     <script>
       window.dataLayer = window.dataLayer || [];
@@ -69,5 +70,6 @@
       gtag('js', new Date());
       gtag('config', '{{ GOOGLE_ANALYTICS }}');
     </script>
+    {% endif %}
   </body>
 </html>


stevejalim:
@atopal If you have access to the GA account, can you prove the relevant JS to drop in, please? (Am assuming it'll gointo `<head>` instead as an async load, but it's been a while since I've had to add it to to a project :smile: )

atopal:
I don't have access to GA yet, so if you can find it, go ahead Steve. I requested access and will follow up once I get it

atopal:
Actually, let's leave the tag manager, it's fine. We can change if necessary after launch

PR #613, opened 2019-10-30
Newrelic agent install

### LINKED ISSUE #521: Monitor Developer Portal
Converted to a separate epic ticket

----

This will be done in two ways:
1) [Speedcurve](https://speedcurve.com) will monitor performance and good practice of the static/published site
2) New Relic APM will monitor the Wagtail CMS for uptime/availability + more as appropriate

----

**Original notes from @limed**

Install the newrelic python agent so that we can start monitoring the application. In order to start monitoring the application we will need to following:

* A new relic license key (@limed)
* Install newrelic agent on Docker image (@limed)
* Generate newrelic config (@limed)
* Start newrelic agent
* Start writing code to measure code performance (@stevejalim)

More info here: https://docs.newrelic.com/docs/agents/python-agent/configuration/python-agent-configuration

limed:
This is an initial creation of newrelic agent install, this will not send the data over to newrelic yet we will need to modify the gunicorn startup code to use the `newrelic-admin` binary. See https://docs.newrelic.com/docs/agents/python-agent/configuration/python-agent-configuration and https://docs.newrelic.com/docs/agents/python-agent/installation/advanced-install-new-relic-python-agent

Once thats done the code itself will need to have further integration, see: https://docs.newrelic.com/docs/agents/python-agent/installation/python-agent-advanced-integration

(Resolves #521)

## Key changes:

- Setup newrelic agent


k8s/Makefile
```diff
@@ -69,12 +76,15 @@ k8s-deployments: k8s-wagtail-deployments k8s-celery-deployments k8s-celery-beat-
...
 k8s-wagtail-deployments:
+	env NEW_RELIC_APP_NAME=dev-portal-web-${TARGET_ENVIRONMENT} \
 	j2 app.deploy.yaml.j2 | ${KC} apply -f -
 
 k8s-celery-deployments:
+	env NEW_RELIC_APP_NAME=dev-portal-celery-${TARGET_ENVIRONMENT} \
 	j2 celery.yaml.j2 | ${KC} apply -f -
 
 k8s-celery-beat-deployments:
+	env NEW_RELIC_APP_NAME=dev-portal-celery-${TARGET_ENVIRONMENT} \
```
escattone: On Kuma, we set this to a blank value since we don't want to track Celery beat (see https://github.com/mdn/infra/blob/8532d61254e5183bd6e5263cd1425e1e336ae099/apps/mdn/mdn-aws/k8s/Makefile#L603). I suspect you would want to do the same? If you do want to track this, I think you'd want a different app name here rather than using the same one as you used for the `k8s-celery-deployments` target above?

limed: Oh good point, fixed this in 0efe30480643d7784c682a4a7733b67cece0c132

etc/newrelic.ini
```diff
@@ -0,0 +1,212 @@
...
+
+# Threshold for query execution time below which query plans
+# will not not be captured. Relevant only when "explain_enabled"
+# is true.
+transaction_tracer.explain_threshold = 0.5
+
+# Space separated list of function or method names in form
+# 'module:function' or 'module:class.function' for which
+# additional function timing instrumentation will be added.
+transaction_tracer.function_trace =
```
stevejalim: @limed should this be blank?

etc/newrelic.ini
```diff
@@ -0,0 +1,212 @@
...
+# The error collector captures information about uncaught
+# exceptions or logged exceptions and sends them to UI for
+# viewing. The error collector is enabled by default. Set this
+# to "false" to turn it off.
+error_collector.enabled = true
+
+# To stop specific errors from reporting to the UI, set this to
+# a space separated list of the Python exception type names to
+# ignore. The exception name should be of the form 'module:class'.
+error_collector.ignore_errors = celery.exceptions:Retry ratelimit.exceptions:Ratelimited django.core.exceptions:DisallowedHost
```
stevejalim: I would prefer we didn't ignore `DisallowedHost` because that's key to spotting if a baked site is healthy (`DisallowedHost` during baking == a site where every HTML page is 0 bytes, but still exists....)

stevejalim: Might also be worth dropping `celery.exceptions:Retry` and `ratelimit.exceptions:Ratelimited` for now, then dropping them back in if we get issues. (We currently don't retry any Celery tasks.)

k8s/app.env.yaml.j2
```diff
@@ -89,4 +103,4 @@
               valueFrom:
                 secretKeyRef:
                   name: dev-portal-secrets
-                  key: OIDC_RP_CLIENT_SECRET
+                  key: OIDC_RP_CLIENT_SECRET
```
stevejalim: Is the missing line of whitespace gonna make a linter choke somewhere?

etc/newrelic.ini
limed: Looks like MDN doesn't have it too, if there is a specific function you want timed we can probably put them here although I believe after we setup the up newrelic decorators  in code we won't need this

k8s/app.env.yaml.j2
limed: Yup this should be fine, apologies my client actually wipes out whitespaces after exit

etc/newrelic.ini
limed: Ok removed in 5ff3c5a77a47324d732b092fa97c1d9549884396

requirements.txt
```diff
@@ -16,6 +16,7 @@ wagtail==2.6.3
 wagtail-bakery==0.3.0
 whitenoise==4.1.4
 urlwait==0.4
+newrelic==5.2.1.129
```
stevejalim: Latest version has already (!) moved on

```suggestion
newrelic==5.2.2.130
```

PR #618, opened 2019-10-31
615: Enable HTTP-only cookies for CSRF

### LINKED ISSUE #615: Try out CSRF `httpOnly` cookies
One last thing from the VA mentioned was trying out `httpOnly` CSRF cookies in the CMS.

Django's docs for the `CSRF_COOKIE_HTTPONLY` setting say:

> If you enable this and need to send the value of the CSRF token with an AJAX request, your JavaScript must pull the value from a hidden CSRF token form input instead of from the cookie.

Talking with the creator of Wagtail, it sounds like Wagtail adheres to that pattern, so this should be low risk / unlikely to break things

stevejalim:
This changeset addresses a request from MozInfoSec to try enabling HTTP-only CSRF cookies.

Talking with `@gasmanic` (Lead Wagtail dev), it sounds like Wagtail doesn't do
anything that should be affected by enabling this, but we should still keep
an eye open.

A light test-drive locally seemed fine and the content-entry phase we're entering will be a good shakedown.

(Resolves #615)

### DIFF
--- a/developerportal/settings/base.py
+++ b/developerportal/settings/base.py
@@ -203,6 +203,7 @@ STATIC_URL = "/static/"
 # Django security settings (see `manage.py check --deploy`)
 
 CSRF_COOKIE_SECURE = True
+CSRF_COOKIE_HTTPONLY = True
 SECURE_BROWSER_XSS_FILTER = True
 SECURE_CONTENT_TYPE_NOSNIFF = True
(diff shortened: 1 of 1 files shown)


stevejalim:
@escattone I notice that this particular value is _not_ set in Kuma. Is there a deliberate reason for that, which might be relevant in Developer Portal?

PR #643, opened 2019-10-31
567: Remove 'Latest authors' section from top of People page

### LINKED ISSUE #567: Remove Latest Authors from people page

stevejalim:
This changeset removes the unit at the top of the People page that showed the three authors with most recently published things.

**BEFORE:**

88dda333-40d2-4464-aee8-caf8788d5ad3

**AFTER** (different data, because rendered locally, and using fallback/placeholder images, which are all unrelated to the changes made here)

Screenshot 2019-10-31 at 22 39 53

The unit is gone, and we've added in a new 'People' `h2`, in the same position as the `h2` on 'Articles'. 

This changeset also drops now-rendant code from People page model that generated that list of latest authors.

(Resolves #567)

### DIFF
--- a/developerportal/apps/people/models.py
+++ b/developerportal/apps/people/models.py
@@ -107,23 +107,6 @@ class People(BasePage):
             "topics": Topic.published_objects.order_by("title"),
         }
 
-    def latest_authors(self, limit=3):
-        """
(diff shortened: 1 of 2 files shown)


stevejalim:
Thanks @escattone !

PR #644, opened 2019-10-31
Increase number of pods for prod

### LINKED ISSUE #523: Increase number of pods for prod
Currently we have 2 worker pods for the CMS site and 1 worker pod of the `celery-worker` on stage. For production push we should increase the number of worker nodes to perhaps 4 and maybe 2 `celery-workers`

All this work is part https://github.com/mdn/developer-portal/issues/466

limed:
Increases the number of `celery-workers` and `web` nodes for the prod
environment.

(Resolves #523)

## Key changes:

- Increase number of pods for celery and web pods

### DIFF
--- a/k8s/config/prod-oregon.sh
+++ b/k8s/config/prod-oregon.sh
@@ -13,7 +13,7 @@ alias kc="kubectl -n ${K8S_NAMESPACE}"
 export APP_SERVICE_CERT_ARN=arn:aws:acm:us-west-2:178589013767:certificate/03cedb62-c36d-4e9e-b5a1-716ca6bdd7c4
 export APP_BUCKET_ROLE_ARN=arn:aws:iam::178589013767:role/developer-portal-prod-us-west-2-role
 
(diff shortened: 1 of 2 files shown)


k8s/config/prod-oregon.sh
```diff
@@ -25,4 +25,6 @@ export APP_AWS_BUCKET_NAME=developer-portal-prod-178589013767
 export APP_AWS_STORAGE_BUCKET_NAME=devportal-media-prod
 export APP_AWS_BUCKET_REGION=us-west-2
 
+export CELERY_WORKER_REPLICAS=2
```
stevejalim: This is probably fine, but we might not need even two workers for ages.

limed: Ok works for me, removed the line which should keep it down to just 1

limed: Fixed in 987c599893d01b2f5d8a55037867bf3747f6caac

PR #652, opened 2019-11-01
625: Allow posts/articles page to also show videos

### LINKED ISSUE #625: Mix blog entries and video on what was "Articles" page and becomes "Posts" page

stevejalim:
This changeset slightly changes the direction of what will soon (via #624) be the Posts page.

It will contain data from Article Pages and also Video Pages. The retrieval of those
was already handled by a helper function, but the HTML templating and JS form-filter
code needed a small update.

Note that while I think this change has no adverse side-effects, there may be a regression somewhere due to the various ways some of the molecule templates are re-used - we should keep a close eye open during content entry.

(Resolves #625)

## Screenshots

Screenshot 2019-11-01 at 15 59 40
Screenshot 2019-11-01 at 15 59 49
Screenshot 2019-11-01 at 15 59 55
Screenshot 2019-11-01 at 16 00 50
Screenshot 2019-11-01 at 16 00 43

## How to test

- add some videos
- give at least some of them topics (via the meta tab)
- view the /articles/ page (soon to be /posts/) and confirm you can see the videos in with the articles
- confirm the filtering works as expected to show/hide the videos and related articles

### DIFF
--- a/developerportal/apps/articles/models.py
+++ b/developerportal/apps/articles/models.py
@@ -267,8 +267,8 @@ class Article(BasePage):
         ]
     )
 
-    # Rss feed
     def get_absolute_url(self):
+        # For the RSS feed
         return self.full_url
 
     @property
--- a/developerportal/apps/articles/models.py
+++ b/developerportal/apps/articles/models.py
@@ -32,7 +32,7 @@ from ..common.blocks import ExternalAuthorBlock, ExternalLinkBlock
 from ..common.constants import RICH_TEXT_FEATURES_SIMPLE
 from ..common.fields import CustomStreamField
 from ..common.models import BasePage
-from ..common.utils import get_combined_articles, get_combined_articles_and_videos
+from ..common.utils import get_combined_articles_and_videos
 
 
 class ArticlesTag(TaggedItemBase):
@@ -105,8 +105,9 @@ class Articles(BasePage):
         return context
 
     @property
-    def articles(self):
-        return get_combined_articles(self)
+    def resources(self):
+        # This Page class will show both Articles/Posts and Videos in its listing
+        return get_combined_articles_and_videos(self)
 
     def get_filters(self):
         from ..topics.models import Topic
--- a/developerportal/apps/articles/templates/articles.html
+++ b/developerportal/apps/articles/templates/articles.html
@@ -19,7 +19,7 @@
           </div>
         </div>
       </div>
-      {% include "organisms/filter-list.html" with type="article" resources=page.articles initial_resources=6 resources_per_page=4 no_resources_message="No posts found" %}
+      {% include "organisms/filter-list.html" with type="article_or_video" resources=page.resources initial_resources=6 resources_per_page=4 no_resources_message="No posts found" %}
     </section>
     {% include "organisms/newsletter-signup.html" %}
   </main>
--- a/developerportal/templates/molecules/cards/card-video.html
+++ b/developerportal/templates/molecules/cards/card-video.html
@@ -8,11 +8,16 @@
 
 <a
   href="{% pageurl resource %}"
-  class="card-link {% if resource.is_external %}js-modal-trigger{% endif %}"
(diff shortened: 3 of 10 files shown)


developerportal/apps/articles/templates/articles.html
```diff
@@ -18,7 +18,7 @@ <h2>Articles</h2>
           </div>
         </div>
       </div>
-      {% include "organisms/filter-list.html" with type="article" resources=page.articles initial_resources=6 resources_per_page=4 no_resources_message="No articles found" %}
+      {% include "organisms/filter-list.html" with type="article_or_video" resources=page.resources initial_resources=6 resources_per_page=4 no_resources_message="No articles found" %}
```
stevejalim: This `article_or_video` approach isn't making me super-happy, but it _is_ explicit, which I prefer to just overriding behaviour when passing `article` around

developerportal/templates/molecules/cards/card-video.html
```diff
@@ -8,12 +8,17 @@
...
   {% if resource.is_external %}
     data-class-name="mzp-has-media"
     data-title="{{ resource.title }}"
   {% endif %}
+  data-type="{{ resource.resource_type }}"
+  {% if filter_target %}
+    data-topics="{% for topic in resource.topics.all %}{{ topic.topic.slug }} {% endfor %}"
+    data-month="{{ resource.month_group|date:'Y-m' }}"
+    hidden
+  {% endif %}
```
stevejalim: This extra markup came from `card-article.html` to ensure both types of content are treated the same way by the filters

developerportal/templates/molecules/cards/card.html
```diff
@@ -5,5 +5,5 @@
 {% elif resource.resource_type == "person" %}
   {% include "molecules/cards/card-person.html" with resource=resource filter_target=filter_target %}
 {% elif resource.resource_type == "video" %}
-  {% include "molecules/cards/card-video.html" with resource=resource %}
+  {% include "molecules/cards/card-video.html" with resource=resource filter_target=filter_target %}
```
stevejalim: I don't thikn adding this extra param is dangerous (re other uses of `card-video.html`) because if `filter_target` is not present, that counts as falsey in Django templates, so the `{% if filter_target %}` blocks added to card-`video.html` will go unevaluated

developerportal/templates/organisms/filter-list.html
```diff
@@ -29,6 +29,12 @@ <h2>{{ heading }}</h2>
               {% for resource in resources %}
                 {% if type == "article" %}
                   {% include "molecules/cards/card.html" with resource=resource.article filter_target=True show_author=True %}
+                {% elif type == "article_or_video" %}
+                  {% if resource.video %}
+                    {% include "molecules/cards/card.html" with resource=resource.video filter_target=True %}
+                  {% else %}
+                    {% include "molecules/cards/card.html" with resource=resource.article filter_target=True show_author=True %}
+                  {% endif %}
```
stevejalim: Again, not the super-nicest approach, but hopefully explicit enough to avoid surprises when we read this back in the future

stevejalim:
Heya @escattone - I know you're at a work-week, but if jetlag means you're awake with time on your hands, could you spare some to look at this PR, please? Cheers!

developerportal/apps/articles/models.py
```diff
@@ -32,7 +32,7 @@
 from ..common.constants import RICH_TEXT_FEATURES_SIMPLE
```
stevejalim: NOTE: feeds.py also needs changing to include mixed content, potentially

stevejalim:
Going to self-review this because everyone else suitable is unavailable

developerportal/templates/molecules/cards/card.html
stevejalim: I still think this 😄

stevejalim:
Tried this locally again and still works. 

Re-reviewed code and it still seems ok.

Will merge and deploy

PR #684, opened 2019-11-05
680: Try another approach to get GA codes rendereed out in static build only

### LINKED ISSUE #680: Tweak selective GA inclusion
While the selective inclusion of GA in the static site only works fine locally, it's not behaving in production (always missing). This needs a fix, naturally 😄

stevejalim:
Locally, when the celery worker runs with `developerportal.settings.worker` (which is the only settings file that tries to read `GOOGLE_ANALYTICS` from the environment), it correctly exports a static site that features the GA code, while Wagtail still leaves the GA code out of the static site.

However, this doesn't appear to be working when deployed to k8s. (And now that staging doesn't have the GA code available in it, we can only see this issue arising in the prod env.)

So, this changeset is assuming that something is stopping the right settings from being used.

By being extra-explicit about which DJANGO_SETTINGS_MODULE to use for the celery worker, we should hopefully make it use the correct settings, without affecting other pods which use the same env-var template

(Resolves #680)

### DIFF
--- a/developerportal/context_processors.py
+++ b/developerportal/context_processors.py
@@ -2,7 +2,12 @@ from django.conf import settings
 
 
 def google_analytics(request):
-    return {"GOOGLE_ANALYTICS": settings.GOOGLE_ANALYTICS}
+    DEFAULT_GA_PLACEHOLDER_VAL = "0"  # from the k8s Makefile
+    output = {}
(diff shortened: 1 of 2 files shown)


stevejalim:
@limed I don't have time to merge this and deploy it to staging right now, but can do later, unless you get to it first -- can do later

PR #699, opened 2019-11-05
Make DJANGO_SETTINGS_MODULE configurable

limed:
This changeset allows `DJANGO_SETTINGS_MODULE` to be a configurable value, this is a companion PR of https://github.com/mdn/developer-portal/pull/684

## Key changes:

- Set a default value `DJANGO_SETTINGS_MODULE` in the `Makefile`
- Export `DJANGO_SETTINGS_MODULE` in the `Makefile` target of your choice to select the value you want


limed:
> Lgtm assuming the first declaration is the default

Correct :)

PR #721, opened 2019-11-06
718: Make it possible to see draft ExternalContent in the Admin

### LINKED ISSUE #718: Display drafts of external content so that admins do not try to add content already present and cause Wagtail to throw a generic 500 error

stevejalim:
This changeset fixes an issue where draft pieces of ExternalContent were not visible in the Admin, which then can cause a server error when the user tries to add the same information again (with the same resulting slug, which causes a 500).

By ensuring the ExternalContent views show both live and draft states, we can avoid this.

Examples of the updated UI:

Screenshot 2019-11-06 at 21 51 12

Screenshot 2019-11-06 at 21 51 20

(Resolves #718)

### DIFF
--- a/developerportal/apps/externalcontent/wagtail_hooks.py
+++ b/developerportal/apps/externalcontent/wagtail_hooks.py
@@ -35,7 +35,32 @@ class ExternalContentButtonHelper(PageButtonHelper):
         return btns
 
 
-class ExternalContentAdmin(ModelAdmin):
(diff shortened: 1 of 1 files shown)


valgrimm:
Still seeing this on prod external videos but maybe a cached issue. That said found workaround of give item different name initially, then save, then change.

stevejalim:
No, this won't be a caching-related issue. Can you provide (via Slack?)
steps to replicate, please?

PR #728, opened 2019-11-07
683: Make it harder to upload a non-SVG for the Topic icon 

### LINKED ISSUE #683: Protect against accidental upload of wrong kind of file for topic icons
An SVG got uploaded as the Topic icon for WebAssembly -- but the field is intended only to _expect_ SVGs, which then basically crashed the homepage (which featured the topic icon) which in turn paged a support engineer to get out of bed.

So, yeah, we should:
a) at least document it clearly in the CMS that it's supposed to be an SVG
b) potentially limit the file-upload options to _only_ SVG if possible

stevejalim:
Add some light protection + help text, to help avoid editors uploading a non-SVG file into the Topic > Theme > Icon slot, which very much needs an SVG file -- if it gets passed a bitmap, then `app_tags.render_svg()` will fail, hard

Includes a migration and simple tests.

----

### Rejection state
Screenshot 2019-11-07 at 10 44 21 1

----

### Success state
Screenshot 2019-11-07 at 11 18 45

(Resolves #683)


developerportal/apps/topics/models.py
```diff
@@ -130,7 +141,16 @@ class Topic(BasePage):
...
-    icon = FileField(upload_to="topics/icons", blank=True, default="")
+    icon = FileField(
+        upload_to="topics/icons",
+        blank=True,
+        default="",
+        help_text=(
+            "MUST be a black-on-transparent SVG icon ONLY, "
+            "with no bitmap embedded in it."
+        ),
+        validators=[check_for_svg_file],
```
Gregoor: Not sure if it helps, but Django ships with a [FileExtensionValidator](https://docs.djangoproject.com/en/2.2/ref/validators/#fileextensionvalidator)

stevejalim: Ah, a 2.1 addition I hadn't noticed - cheers!

PR #734, opened 2019-11-07
705: Tune logging config

### LINKED ISSUE #705: Suppress "invalid HTTP_HOST" warnings from bots/spiders, etc
Since we went into production, @limed and I have received hundreds of emails related to requests to the Prod CMS's IP with fake/unexpected hostnnames in the header. (ie, they are not in [`ALLOWED_HOSTS`](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-ALLOWED_HOSTS))

All/most of these are likely probes/bots - eg http://mehmet.onyphe.io/ is definitely some of the requests.

The hundreds of emails create a lot of noise, making it harder to spot any genuinely pressing issues with Prod.

So, am wondering about just ignoring the alerts when we get a fake/bad hostname in prod and staging, via this approach: https://docs.djangoproject.com/en/2.2/topics/logging/#django-security

stevejalim:
This changeset addresses the needs to make logging config more explicit AND silence a single, particular kind of logged warning that's saturating the inboxes of `settings.ADMINS`: `django.core.exceptions.DisallowedHost`

## Key changes:

* Rather than extend `DEFAULT_LOGGING` in our `settings.base`, show an explicit `LOGGING` config (based initially on the content of `DEFAULT_LOGGING`)
* Add a `null` handler / blackhole 
* With the blessing of @limed, use that `null` handler to trap all logger calls for `DisallowedHost`, which are drowning us. There's precedent for this approach, including [the official docs](https://docs.djangoproject.com/en/2.2/topics/logging/#django-security) and other, solid, Django sites like [readthedocs](https://github.com/readthedocs/readthedocs.org/pull/4879/files).

Note that blackholing `DisallowedHost` error logging doesn't stop the actual blocking, it just stops it appearing in logs (including triggering the `mail_admins` handler)

(Resolves #705)

### DIFF
--- a/developerportal/settings/base.py
+++ b/developerportal/settings/base.py
@@ -13,7 +13,6 @@ https://docs.djangoproject.com/en/2.1/ref/settings/
 import os
 
 from django.core.management.utils import get_random_secret_key
-from django.utils.log import DEFAULT_LOGGING
 
 from wagtail.embeds.oembed_providers import all_providers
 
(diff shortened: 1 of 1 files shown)


developerportal/settings/base.py
```diff
@@ -367,8 +366,52 @@
...
+            "level": "INFO",
+            "class": "logging.StreamHandler",
+            "formatter": "django.server",
+        },
+        "mail_admins": {
+            "level": "ERROR",
+            "filters": ["require_debug_false"],
+            "class": "django.utils.log.AdminEmailHandler",
+        },
+        "null": {"class": "logging.NullHandler"},
```
stevejalim: This line is an addition over `DEFAULT_LOGGING`

developerportal/settings/base.py
```diff
@@ -367,8 +366,52 @@
...
+    },
+    "loggers": {
+        "django": {"handlers": ["console", "mail_admins"], "level": "INFO"},
+        "django.server": {
+            "handlers": ["django.server"],
+            "level": "INFO",
+            "propagate": False,
+        },
+        "mozilla_django_oidc": {"handlers": ["console"], "level": "INFO"},
```
stevejalim: This `mozilla_django_oidc` line is an addition over `DEFAULT_LOGGING`, as is everything below it

PR #735, opened 2019-11-08
726 Surface hidden slug clashes for external content

### LINKED ISSUE #726: External content slugs should be specific to post or video so that title conflicts are avoided
Tangentially related to 721

When an external post and an external video have the same title, they clash (example: "Why is CSS So Weird?". Make the [hidden] slug a bit smarter to adjust for this.

NOT a blocker, Val has a workaround to name the new video/post differently when a clash identified, then go back and fix the title.

stevejalim:
This changeset addresses a blow-up when adding External entities (any subtype) that has a name that auto-creates a slug that matches an existing slug for another piece of `External[Content|Video|Article|Event]`.

(For context an `External[Content|Video|Article|Event]` is still a subclass of a Wagtail `Page`, but when it is used we never render it as a regular Wagtail page: we just use it as a way of holding data about an external URL to load.) 

Prior to this changeset, because the `slug` field for the entity was not surfaced in Wagtail's Admin UI, there was no way to catch and report a slug-field clash as a `ValidationError`, so the conflict would cause an integrity-related blow-up during `save()`. (The slug field was probably not shown, because an `External*` item isn't meant to be used as a browsable page in the rendered site.)

The fix has been to reveal the `slug` in the Admin UI, to be able to catch and fix the clashes, while also adding useful help-text (for `External*` only) for editors.

(Resolves #726)

## Screenshots

**Note the presence of the `slug` field, the error messaging and also the help text (to the right)**
Screenshot 2019-11-08 at 11 53 33

**Note that non-`External*` entities still get the existing, dynamic, help text**
Screenshot 2019-11-08 at 11 55 28

### DIFF
--- a/developerportal/apps/common/forms.py
+++ b/developerportal/apps/common/forms.py
@@ -1,17 +1,34 @@
 from wagtail.admin.forms import WagtailAdminPageForm
 from wagtail.core.models import Site
 
+EXTERNAL_ENTITY_NAMES = [
+    # Done as strings to avoid cross-module imports
+    "ExternalArticle",
+    "ExternalContent",
+    "ExternalEvent",
+    "ExternalVideo",
+]
 
-def _custom_slug_help_text():
-    # Generate slug help_text that uses the default site’s real URL, falling
-    # back to example.com instead of the Wagtail default.
-    default_site = Site.objects.filter(is_default_site=True).first()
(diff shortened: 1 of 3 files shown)


stevejalim:
@Gregoor yeah, no worries - it is a bit tricky to set up state. I should include notes for such things in future PRs now that it's not just me involved 😄

PR #755, opened 2019-11-11
Set GID and UID for cfssl container

### LINKED ISSUE #753: Fix cert permissions generated by gen-cert.sh
The `gen-cert.sh` script generates a CA and certs for local development, however when ran in Jenkins the cert it generates is owned by `root` so Jenkins throws the following error

```
error checking context: 'no permission to read from '/var/lib/jenkins/workspace/developer-portal_master/etc/nginx/ssl/ca-key.pem''.
```

We need to make sure that the certs that i generates is owned by the `jenkins` user and group

limed:
Ensures that GID and UID is set for the `cfssl` container, this ensure that Jenkins eventually owns the cert files

(Resolves #753)

## Key changes:

- Sets `GID` and `UID`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,6 +77,9 @@ node {
         }
       }
       stage ('Test') {
+        // Set UID and GID
+        env['UID'] = sh(returnStdout: true, script: 'id -u jenkins').trim()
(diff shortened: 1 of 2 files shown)


limed:
r=limed

PR #756, opened 2019-11-11
Set NEW_RELIC_APP_NAME to emptry string in migration job

limed:
Set `NEW_RELIC_APP_NAME` to emptry string for db migration job since it doesn't need it.


limed:
r=limed

Issue #781, opened 2019-11-15
Standardise SVG width for topic icons


stevejalim:
Slightly different-width icons mean that the content in the Topics panels on the homepage flow slightly differently.

Might be worth me downloading them, resizing them to have a consistent width and re-uploading them.

At the same time, I can do a Voice SVG if someone points me at a source icon

Example screenshot:

![maccat_firefox_62.0.jpg](https://images.zenhubusercontent.com/5d7b94dd167772000187fd33/5bc3c89e-3909-4a1e-a15b-0831a225dadf)

stevejalim:
So, tinkering with CSS I can improve it to a certain point, but this looks like it's something that's best addressed by ensuring all the SVG icons are the same width and centred vertically and horizontally in their own files/definitions, because setting the CSS to fix the width of the icon panels and centre-aligning them does help, but also highlights how offset the "top" of some icons are.

eg: 

![Screenshot 2019-11-16 at 13.22.19.png](https://images.zenhubusercontent.com/5d7b94dd167772000187fd33/d2ea84a1-5df1-41a6-ae3a-be1b8fd9a7ab)


What do you think of the above @atopal - "good enough"? I am happy to try the additional step of re-cutting the SVG icons on Monday, no worries.

atopal:
Yes, absolutely good enough, Steve :)

PR #794, opened 2019-11-18
781: Minor CSS changes to Topic blocks to better align the rich text

### LINKED ISSUE #718: Display drafts of external content so that admins do not try to add content already present and cause Wagtail to throw a generic 500 error

stevejalim:
While this isn't a perfect fix, it makes the content less susceptible to reflow issues triggered by differing widths of the SVG icons for each Topic. Note that the existing SVG icons do not need to be changed for this to be "good enough" (See comments in #781), but standardising them will also help.

Checked locally and/or in Browserstack across Edge, FF, Safari and Chrome

## Before
Screenshot 2019-11-18 at 11 24 28

## After
Screenshot 2019-11-18 at 11 24 10

(Resolves #718)

### DIFF
--- a/src/css/molecules/item-topic-link.scss
+++ b/src/css/molecules/item-topic-link.scss
@@ -47,6 +47,8 @@
   font-size: 0; // Prevent spacing below icon.
   margin-right: 16px;
   padding-top: 4px;
+  width: 50px;
+  text-align: center;
 
   img {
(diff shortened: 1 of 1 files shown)


stevejalim:
@atopal I know you can't formally code review this, but a 👍 would be welcome (Gregor is unavailable today)

Issue #796, opened 2019-11-18
CRITICAL: production celery is crashing during static site build


stevejalim:
Something is killing the celery worker when building a static site:

```
[2019-11-18 12:04:02,969: INFO/MainProcess] Received task: developerportal.apps.staticbuild.wagtail_hooks._static_build_async[b78c5c08-3190-4e71-b139-3b1f7c3f0783]
[2019-11-18 12:04:02,974: INFO/ForkPoolWorker-89] [Static-build-and-sync] Attempting fresh build
[2019-11-18 12:04:02,975: INFO/ForkPoolWorker-89] [Static-build-and-sync] Created temporary build dir /app/build/2019-11-18T12:04:02.975426+00:00
[2019-11-18 12:04:02,975: INFO/ForkPoolWorker-89] [Static-build-and-sync] (wagtail-bakery) command 'Build' started.
[2019-11-18 12:04:03,021: INFO/ForkPoolWorker-89] Build started
[2019-11-18 12:04:17,340: ERROR/MainProcess] Process 'ForkPoolWorker-89' pid:15428 exited with 'signal 9 (SIGKILL)'
[2019-11-18 12:04:17,356: ERROR/MainProcess] Task handler raised error: WorkerLostError('Worker exited prematurely: signal 9 (SIGKILL).')
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/billiard/pool.py", line 1267, in mark_as_worker_lost
    human_status(exitcode)),
billiard.exceptions.WorkerLostError: Worker exited prematurely: signal 9 (SIGKILL).
2019-11-18 12:04:17,427 (15498/MainThread) newrelic.core.agent INFO - New Relic Python Agent (5.2.3.131)
[2019-11-18 12:04:17,427: INFO/ForkPoolWorker-90] New Relic Python Agent (5.2.3.131)
2019-11-18 12:04:18,788 (15498/NR-Activate-Session/dev-portal-celery-prod) newrelic.core.data_collector INFO - Successfully registered New Relic Python agent where app_name='dev-portal-celery-prod', pid=15498, redirect_host='collector-005.newrelic.com' and agent_run_id='REDACTED', in 1.34 seconds.
[...]

PR #799, opened 2019-11-18
796: Attempt to fix the crashing celery worker by boosting RAM

stevejalim:
This changeset hopefully addresses the dying celery workers in #796 boosting RAM 4x: from 128Mi to 512Mi default and 256Mi to 1024Mi limit


stevejalim:
@limed I don't know if this will work, but assuming it does, are we OK in terms of real-world resource spend etc?

PR #829, opened 2019-11-24
746: Add sitemap support to static site

### LINKED ISSUE #746: Add sitemap support
We don't currently export one to the static site.

AC
- [x] Code in production
- [x] Search console show no errors for sitemap
- [x] Search console reports all pages as indexed (should be about 171)

stevejalim:
This changeset adds an XML sitemap to the rendered site and also includes it in the static site exported to S3

(Addresses #746)

## How to test

- locally, run `docker-compose exec app manage.py build` to "bake out" the static site to `/app/build/` in your `app_1` container
- shell in with `docker-compose exec app sh` 
- `less /app/build/index.html` to see the link to `sitemap.xml` in the `<head>` of the document
- `less /app/build/sitemap.xml` to see the sitemap content itself.

For convenience, I've temporarily exposed my own dev S3 bucket in website mode, so you can check for yourself. The content is dummy, but the sitemap is still correct. (Note that there is no HTTPS on this example bucket)

- View source on http://devportalstatictestxyz.s3-website-eu-west-1.amazonaws.com/ - see that the sitemap is linked from the `head`
- Browse the sitemap itself at http://devportalstatictestxyz.s3-website-eu-west-1.amazonaws.com/sitemap.xml and confirm it works fine

## To check on staging after merge:

- [ ] Ensure HTTPS links are correctly included in the sitemap

### DIFF
--- a/developerportal/apps/bakery/tests/test_views.py
+++ b/developerportal/apps/bakery/tests/test_views.py
@@ -267,3 +267,13 @@ class CloudfrontInvalidationViewTests(TestCase):
                 )
             ],
         )
+
+
+class SitemapViewTests(TestCase):
+    def test_reminder(self):
+        self.fail("WRITE ME")
+
+
+class HelpersTests(TestCase):
+    def test_is_secure_request(self):
+        self.fail("WRITE ME")
--- a/developerportal/apps/bakery/tests/test_views.py
+++ b/developerportal/apps/bakery/tests/test_views.py
@@ -3,6 +3,7 @@
 import datetime
 from unittest import mock
 
+from django.conf import settings
 from django.test import TestCase, override_settings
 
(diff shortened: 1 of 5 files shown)


atopal:
okay, I can see the portal in search console now. Only 88 pages were indexed by Google. I added the sitemap and it discovered 224 pages. No errors. 

Wagtail says that we have 312 pages, but I'm not sure how to see how many of those are actually published pages meant to be indexed.

stevejalim:
The root cause of the difference is, I think, that external content (links,
events, videos) were modelled as pages but are never rendered as pages.

As such, the total in the site map should be:
total_pages - total_items_listed_as_external_content

I can check the exact numbers when .at a real screen but the totals you
mentioned sound right.

atopal:
In that case let's go ahead and close this

Issue #865, opened 2019-11-25
Ensure ExternalContent subclasses are made as child pages of the relevant parent page


stevejalim:
While dealing with #864 I noticed that none of the subclasses of `ExternalContent` has `parent_page_types` set. This explains (I think, and would need to be 100% sure) why they are made as children of the _unreachable_ Root page, and return `None` for `page.get_site()`.

While this can be worked around, it's better to clean things up:

- [ ] Ensure each `ExternalArticle` has `Articles` in its `parent_page_types`  
- [ ] Ensure each `ExternalEvent` has `Events` in its `parent_page_types`  
- [ ] Ensure each `ExternalVideo` has `Videos` in its `parent_page_types` 
- [ ] Add a data migration to non-destructively bring this into line for existing pages
- [ ] Update documentation around the need to use the default site, added in #864

**BEFORE TACKLING THIS, ENSURE #868 IS UNDERSTOOD PROPERLY**

stevejalim:
@liamnewmarch @dcgauld No worries if you can't answer, but does the above issue/suggestion/approach make sense to you, or is there a reason `parent_page_types` wasn't set that I am not appreciating? Cheers!

liamnewmarch:
I’m not sure I fully understand, but in not setting `parent_page_types` this should allow all pages to be parent [by default][1]. Being more explicit here does sound like it would prevent ExternalContent from being created where it shouldn’t, so the theory sounds good to me.

[1]: https://github.com/wagtail/wagtail/blob/96965d/wagtail/core/models.py#L919-L922

PR #867, opened 2019-11-25
864: Fix regression with static-site generation

### LINKED ISSUE #864: ExternalContent subclasses blows up latest wagtail-bakery release
Not sure if it's a `wagtail-bakery` regression or just an awkward gotcha of the `externalcontent` app but since `wagtail-bakery==0.4.0` hit `master`, static builds are broken because the `.get_site()` returns None on all `ExternalContent` subclasses (see #865 ) and wagtail-bakery (or django-bakery) doesn't like that, throwing a `ValueError`.

Once quick fix will be to simply return the default `Site`, given we only have one use anyway, while #865 will improve things and remove the need for a workaround

stevejalim:
Somewhere between wagtail-bakery 0.3 and 0.4 (which was recently bumped as a dependency) a change took place that surfaces a shortcoming in our code: ExternalContent subclasses return None for .get_site() because they are not saved as children of Articles, Videos or Events, respectively.

The workaround for now is to just get the default Site for these Pages. A fuller fixup has been ticketed as #865.

Note that bumping django-bakery in this changeset didn't cause the regression - it was happening before that, and this change is just to bring it up to date, now that wagtail-bakery==0.4.0 allows a higher version of django-bakery.

I've also added a reminder to the developer notes about the downside of assuming we can always use the default Site if we lack a route to a Page's actual Site.

Manually checked a static build (all the way to S3) and seems fine:  http://devportalstatictestabc.s3-website-eu-west-1.amazonaws.com

Python tests pass; JS tests/linting checks pass

(Resolves #864)

### DIFF
--- a/developerportal/apps/bakery/views.py
+++ b/developerportal/apps/bakery/views.py
@@ -34,7 +34,7 @@ class AllPublishedPagesViewAllowingSecureRedirect(AllPublishedPagesView):
         Build wagtail page and set SERVER_NAME to retrieve corresponding site
         object.
         """
-        site = obj.get_site()
+        site = obj.specific.get_site()
         logger.debug("Building %s" % obj)
         secure_request = site.port == 443 or getattr(
             settings, "SECURE_SSL_REDIRECT", False
--- a/developerportal/apps/externalcontent/models.py
+++ b/developerportal/apps/externalcontent/models.py
@@ -99,7 +99,7 @@ class ExternalContent(BasePage):
         return self.external_url
 
     def get_site(self):
-        """Due to modelling/config (which will be changed soon), it's possiblethat
+        """Due to modelling/config (which will be changed soon), it's possible that
         ExternalContent subclasses end up a children of Wagtail's hidden root page,
         rather than the Homepage, which means that the default get_site() will not
         work for them and instead will return None. This addresses that
--- a/developerportal/apps/externalcontent/models.py
+++ b/developerportal/apps/externalcontent/models.py
@@ -100,7 +100,7 @@ class ExternalContent(BasePage):
 
     def get_site(self):
         """Due to modelling/config (which will be changed soon), it's possiblethat
-        ExternalContent subclasses end up being children of Wagtail's root page,
+        ExternalContent subclasses end up a children of Wagtail's hidden root page,
         rather than the Homepage, which means that the default get_site() will not
         work for them and instead will return None. This addresses that
         shortcoming by fetching the default Site instead, if needed."""
--- a/developerportal/apps/externalcontent/models.py
+++ b/developerportal/apps/externalcontent/models.py
@@ -1,5 +1,6 @@
 # pylint: disable=no-member
 import datetime
+import logging
 
 from django.db.models import (
     CASCADE,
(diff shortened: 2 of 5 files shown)


developerportal/apps/bakery/views.py
```diff
@@ -34,7 +34,7 @@ def build_object(self, obj):
         Build wagtail page and set SERVER_NAME to retrieve corresponding site
         object.
         """
-        site = obj.get_site()
+        site = obj.specific.get_site()
```
stevejalim: `.specific` lets us call `get_site()` on our Page subclass so we can use our fallback solution

developerportal/apps/externalcontent/models.py
```diff
@@ -95,6 +98,19 @@ def get_url(self, request=None, current_site=None):
     def relative_url(self, current_site, request=None):
         return self.external_url
 
+    def get_site(self):
+        """Due to modelling/config (which will be changed soon), it's possiblethat
```
stevejalim:
```suggestion
        """Due to modelling/config (which will be changed soon), it's possible that
```

developerportal/apps/externalcontent/models.py
```diff
@@ -95,6 +98,19 @@ def get_url(self, request=None, current_site=None):
...
+        ExternalContent subclasses end up being children of Wagtail's root page,
```
stevejalim:
```suggestion
        ExternalContent subclasses end up a children of Wagtail's hidden root page,
```

docs/developer-notes.md
```diff
@@ -11,3 +11,8 @@ As a result, the changes done were all kept (as part of ticket 624) at surface l
 Streamfield blocks have been renamed from 'article' to 'post' to keep the UI consistent, but the internal use of a `type` or `resource_type` attribute on a Page instance has been left with the value 'article', not changed to 'post' because that would have been a non-visible-to-humans change.
 
 However, in the future, we may decide that we definitely want to stick with "Posts", in which case a proper model-renaming exercise is worth it, along with CSS and template-name changes, too.
+
+## General assumption about there being just one Site
+
+Sometimes in the custom views that inform the 'baking' process for wagtail-bakery, we've had to get hold of a Site object from the database instead of via a relation from a Page we're processing.
+In this case, we've had no option but to use the default Site -- which is fine because was only one at the genesis of this project. HOWEVER, if this changes and - for example - we add i18n and l10n, this assumption will no longer be reliable/safe and any ORM query that uses `is_default_site=` must be refactored out of the codebase.
```
stevejalim:
```suggestion
In this case, we've had no option but to use the default Site -- which was initially fine because there was only one at the genesis of this project. HOWEVER, if this changes and - for example - we add i18n and l10n, this assumption will no longer be reliable/safe and any ORM query that uses `is_default_site=` must be refactored out of the codebase.
```

src/js/utils.js
```diff
@@ -21,6 +21,7 @@ exports.parseQueryParams = () => {
 /**
  * Creates an object based on a form's current input values.
  *
+ * @param {object} form
```
stevejalim: This is a minor cleanup spotted by the newer eslint-jsdoc plugin, which was bumped on `master` earlier today

PR #899, opened 2019-11-28
803: Add Sentry support

### LINKED ISSUE #803: Send stacktraces to sentry
We need to start sending out stacktraces or any errors to [sentry](https://sentry.io/)

Things to do:
- [x] Create new sentry project
- [x] Create new sentry key
- [ ] Expose `SENTRY_DSN` env var
- [ ] Update developer-portal code to start capturing errors

The first 2 tasks help from cloudops and the third tasks will require work from @stevejalim

stevejalim:
Automatically enabled when SENTRY_DSN is available as an env var

Supoprts the idea of a REVISION_HASH as per Kuma, but I don't know if it's available (but it would be nice to add)

(Resolves #803)

Tested locally using staging credentials, sending from both the test management command and by raising an exception in a Celery task (to prove we'll get alerts from the worker, too):

Screenshot 2019-11-28 at 16 54 26

### DIFF
--- a/developerportal/settings/base.py
+++ b/developerportal/settings/base.py
@@ -71,6 +71,8 @@ INSTALLED_APPS = [
     "django_celery_results",
     "mozilla_django_oidc",  # needs to be loaded after auth
 ]
+# Note that "raven.contrib.django.raven_compat" is added to INSTALLED_APPS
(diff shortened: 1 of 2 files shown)


stevejalim:
Deployed to staging and production and sent test messages there - all fine

PR #918, opened 2019-12-02
906: Switch from Google Tag Manager to regular GA; add in Mozilla DNT support

### LINKED ISSUE #906: Move Google Analytics code to the <head>
cc @atopal for visibility

AC 
- [x] Regular Google Analytics is loaded, not GTM
- [x] Mozilla DNT code / helper is in use, as per Kuma

stevejalim:
This changeset moves away from using GTM and instead drops in GA code in the head of the document, as per mdn/kuma

At the same time it also uses the same Do Not Track lib [that kuma uses](https://github.com/mdn/kuma/blob/master/kuma/static/js/libs/mozilla.dnthelper.js) - thanks @sneethling!

Note that the DNT lib lifted from kuma is slighty changed to pin a `Mozilla` object on the DOM's `window` so that it's within reachable scope within the page. The way Kuma integrates the same code is to inline it into the HTML (ie, render it as a template) but that behaviour isn't currently supported by the Developer Portal codebase, and switching to it seemed more disruptive than just updating webpack to output a new bundle.

As before, if there is no `GOOGLE_ANALYTICS` code present via settings, the relevant JS code block is not included. If it is present, the block is rendered. As such, only `settings.worker` will, by default, try to get `GOOGLE_ANALYTICS` from the env vars, so only the static site should render it out.

Example of included HTML/JS when GA is meant to be rendered (fake code here is `TestTestTest`

```
  <!-- Mozilla DNT Helper -->
  <script src="/static/js/head.4da485bfc790.js"></script>

  <!-- Only load GA if DNT is not enabled -->
  <script>
  if (window.Mozilla && !window.Mozilla.dntEnabled()) {
      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
      })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

      ga('create', 'TestTestTest', 'mozilla.com');
      ga('set', 'anonymizeIp', true);
      (function() {
          // Remove analytics clutter from campaign URLs
          var win = window;
          var removeUtms = function(){
              var location = win.location;
              if (location.href.indexOf('utm') != -1 && win.history.replaceState) {
                  win.history.replaceState({}, '', location.pathname);
              }
          };
          ga('send', 'pageview', {'hitCallback': removeUtms});
      })();
  }
  </script>
```

(Resolves #906)

# Testing
Requires a local version of the stack and an S3 bucket configured appropriately. If you lack these things, a code-only review is fine. 

If you do have devportal running locally:
* ensure `GOOGLE_ANALYTICS` is defined in your `.env` but not your `local.py` 
* View source on `localhost:8000` and confirm there is no GA code/JS rendered 
* Run `docker-compose exec app python manage.py request_static_build` and view source on the index.html that ends up in your S3 bucket - that _will_ show the GA code/JS

### DIFF
--- a/src/js/mozilla-dnthelper.js
+++ b/src/js/mozilla-dnthelper.js
@@ -1,0 +1,65 @@
+/* eslint-disable no-octal-escape */
+/* eslint-disable func-names */
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// create namespace
+if (typeof window.Mozilla === 'undefined') {
+  window.Mozilla = {};
+}
+
+/*
+ * Returns true or false based on whether doNotTack is enabled. It also takes into account the
+ * anomalies, such as !bugzilla 887703, which effect versions of Fx 31 and lower. It also handles
+ * IE versions on Windows 7, 8 and 8.1, where the DNT implementation does not honor the spec.
+ * @see https://bugzilla.mozilla.org/show_bug.cgi?id=1217896 for more details
+ * @params {string} [dnt] - An optional mock doNotTrack string to ease unit testing.
+ * @params {string} [ua] - An optional mock userAgent string to ease unit testing.
+ * @returns {boolean} true if enabled else false
+ */
+window.Mozilla.dntEnabled = function(dnt, ua) {
+  // for old version of IE we need to use the msDoNotTrack property of navigator
+  // on newer versions, and newer platforms, this is doNotTrack but, on the window object
+  // Safari also exposes the property on the window object.
+  let dntStatus =
+    dnt || navigator.doNotTrack || window.doNotTrack || navigator.msDoNotTrack;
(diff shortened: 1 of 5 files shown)


stevejalim:
CI currently failing:

`Module not found: Error: Can't resolve './lib/mozilla.dnthelper' in '/app/src/js'`

Will explore

src/js/mozilla-dnthelper.js
```diff
@@ -0,0 +1,65 @@
+/* eslint-disable no-octal-escape */
```
stevejalim: This code has been updated from the kuma version to make eslint happier, plus also to pin the `Mozilla` object on window, to make it reachable from the template. 

Please shout if you think this isn't the right way to go

schalkneethling:
lgtm :+1:

stevejalim:
Given @schalkneethling has provided a 👍 am going to merge this and release it

PR #926, opened 2019-12-03
886: Add template filter to move the cachebreaker string from inside the filename to a querystring

### LINKED ISSUE #868: Explore/address issues with upgrading to wagtail-bakery 0.4.0
For a fairly innocuous-seeming version bump, `wagtail-bakery==0.4.0` breaks our static build because it's trying to build pages from the `externalcontent` app that are not intended to be rendered, and have no Site associated with them. 

ie, `ExternalVideo`, `ExternalEvent` and`ExternalArticle` inherit `Page` and have 'cards' so they can have summaries rendered in list views, etc, but they are never rendered as full pages - they only ever have their `external_url` value used.

* wagtail-bakery 0.3.0 does NOT render out those `External*` pages 

* wagtail-bakery 0.4.0 does render them out, for some reason, and then blows up because there is no `Site` related to the page (rightly or wrongly - that doesn't matter right now)

And here's "proof" that it's `wagtail-bakery==0.4.0` that trips up our issue, not `django-bakery` because `0.12.3` is what we've been using all along

```
------------------------------------------------------------------------
wagtail-bakery==0.4.0
  - django-bakery [required: ~=0.12.7, installed: 0.12.3]
/app $ python manage.py build
INFO:bakery.management.commands.build:Build started
  File "manage.py", line 14, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.7/site-packages/bakery/management/commands/build.py", line 112, in handle
    self.build_views()
  File "/usr/local/lib/python3.7/site-packages/bakery/management/commands/build.py", line 241, in build_views
    self.get_view_instance(view).build_method()
  File "/usr/local/lib/python3.7/site-packages/wagtailbakery/views.py", line 101, in build_queryset
    self.build_object(item)
  File "/app/developerportal/apps/bakery/views.py", line 39, in build_object
    secure_request = site.port == 443 or getattr(
AttributeError: 'NoneType' object has no attribute 'port'
```

```
------------------------------------------------------------------------
wagtail-bakery==0.3.0
  - django-bakery [required: ==0.12.3, installed: 0.12.3]
/app $ python manage.py build
INFO:bakery.management.commands.build:Build started
INFO:bakery.management.commands.build:Build finished
```

--

In addition, the workaround I put in (#864) to fall back to using the default Site ends up triggering very odd warnings on stage during an otherwise correct-seeming build process, so has been backed out

This is from the 'fixed' workarond from #864 where we fall back to the default Site if need be:
```
[2019-11-25 23:00:59,415: INFO/ForkPoolWorker-2] Page WebAssembly's post-MVP future: A cartoon skill tree lacks a Site. Using default Site...
[2019-11-25 23:00:59,485: WARNING/ForkPoolWorker-2] Not Found: /2018/10/webassemblys-post-mvp-future/
[2019-11-25 23:00:59,488: INFO/ForkPoolWorker-2] Page Technical Details on the Recent Firefox Add-on Outage lacks a Site. Using default Site...
[2019-11-25 23:00:59,544: WARNING/ForkPoolWorker-2] Not Found: /2019/05/technical-details-on-the-recent-firefox-add-on-outage/
[2019-11-25 23:00:59,548: INFO/ForkPoolWorker-2] Page Add-Ons Outage Post-Mortem Result lacks a Site. Using default Site...
[2019-11-25 23:00:59,605: WARNING/ForkPoolWorker-2] Not Found: /2019/07/add-ons-outage-post-mortem-result/
[2019-11-25 23:00:59,608: INFO/ForkPoolWorker-2] Page Extensions in Firefox 69 lacks a Site. Using default Site...
[2019-11-25 23:00:59,668: WARNING/ForkPoolWorker-2] Not Found: /addons/2019/08/05/extensions-in-firefox-69/
```
Two strange things that need solving to go forward:
- [ ] those pages in INFO _do_ have Sites available to them via parent pages
- [ ] the Not Found warnings make no sense - they appear to be referencing URLs that are in the _content_ of the rendered page. This might just be how the baking process works and it's now that we have logging imported that it's showing up something that was otherwise not being surfaced, but _that_ seems odd too

--

# Extra info to think about
- 0.3.0 doesn't try to build out subclasses of ExternalContent, but 0.4.0 does... That's a/the reason why it doesn't explode even though the `External*` pages lack a proper parent... The answer for why that happens is surely in https://github.com/wagtail/wagtail-bakery/compare/0.3.0...0.4.0 somewhere...?

stevejalim:
This changeset adds and uses a new template filter that, for a given URL, checks if it has a cachebreaking hash and if it does, converts it to be a URL that uses the hash as a querystring cachebreaker instead.

eg:
    INPUT: 'https://example.com/static/foo.12313abc.jpg'
    OUTPUT: 'https://example.com/static/foo.jpg?h=12313abc'

Why? So that images automatically embedded in social media posts (via OpenGraph tags) don't end up as broken links with a new release that changes the hash in the filename -- by moving the hash to a querystring, we get the benefit of cachebreaking but with a much softer edge that won't result in links 404ing even if the cachebreaking hash has changed. (ie, querystring cachebreakers are more forgiving.)

Note that `ManifestFileStorage` (used by `django-whitenoise` in this project) makes hash-named versions of the files, but also _does_ keep the original around,so we are safe to assume that if `static/img/foo.213123.jpg` exists, then `static/img/foo.jpg` will, too

This template filter is _only_ used for in `social-meta.html`, for the OpenGraph image tag 

(Resolves #868)

## Before

Screenshot 2019-12-03 at 21 28 26

## After

Screenshot 2019-12-03 at 21 04 34

## Proof that images not served from staticfiles are not affected

Screenshot 2019-12-03 at 21 05 26

And here's proof that the default placeholder image (which is the common case here) will be loaded regardless of the querystring:
* https://developer.mozilla.com/static/img/placeholders/article.jpg?h=0asdasda
* https://developer.mozilla.com/static/img/placeholders/article.jpg
* https://developer.mozilla.com/static/img/placeholders/article.jpg?h=differentquerystring

### DIFF
--- a/developerportal/templatetags/app_tags.py
+++ b/developerportal/templatetags/app_tags.py
@@ -1,5 +1,7 @@
 import binascii
+import logging
 import os
+import re
 from mimetypes import guess_type
 
 from django import template
@@ -7,6 +9,8 @@ from django.conf import settings
 from django.utils.safestring import mark_safe
 
 register = template.Library()
+
+logger = logging.getLogger(__name__)
 
 
 @register.simple_tag
@@ -57,3 +61,38 @@ def has_at_least_two_filters(filters):
     # least two of its keys' values are truthy, else False
     result = len([val for val in filters.values() if bool(val)]) >= 2
     return result
+
+
+@register.filter
+def filename_cachebreaker_to_querystring(url):
+    """For the given URL, check if it has a cachebreaking hash. If it does,
+    convert it to be a URLthat uses the hash as a querystring cachebreaker
+    instead.
+
+    eg
+        INPUT: 'https://example.com/static/foo.12313abc.jpg'
+        OUTPUT: 'https://example.com/static/foo.jpg?h=12313abc'
+
+    Why? So that images automatically embedded in social media posts (via OG
+    tags) don't end up as broken links with a new release that changes the hash
+    in the filename -- by moving the hash to a querystring, we get the benefit
+    of cachebreaking but with a much softer edge that won't result in links
+    404ing even if the cachebreaking hash has changed. (ie, querystring
+    cachebreakers are more forgiving.)
+
+    Note that ManifestFileStorage (used by django-whitenoise in this project) makes
+    hash-named versions of the files, but also _does_ keep the original around,
+    so we are safe to assume that if static/img/foo.213123.jpg exists, then
+    static/img/foo.jpg will, too
+    """
+
+    pattern = re.compile(r"(\.[a-f0-9]+)(\.\w+)$")
+    hits = pattern.search(url)
+    if not hits:
+        logger.info(f"Couldn't extract has from URL {url}. Leaving unchanged.")
+        return url
+    dotted_hash = hits.groups()[0]
+    url = url.replace(dotted_hash, "")
+    _hash = dotted_hash[1:]  # the [1:] skips the . at the start
+    url = f"{url}?h={_hash}"
+    return url
--- a/developerportal/templatetags/app_tags.py
+++ b/developerportal/templatetags/app_tags.py
@@ -86,7 +86,7 @@ def filename_cachebreaker_to_querystring(url):
     static/img/foo.jpg will, too
     """
 
-    pattern = re.compile(r"(\.[a-f0-9]+)(\.\w+)$")
+    pattern = re.compile(r"(\.[a-f0-9]+)\.\w+$")
     hits = pattern.search(url)
     if not hits:
         logger.info(f"Couldn't extract has from URL {url}. Leaving unchanged.")
--- a/developerportal/apps/common/tests/test_app_tags.py
+++ b/developerportal/apps/common/tests/test_app_tags.py
@@ -2,7 +2,10 @@ from django.test import TestCase
 
 from wagtail.core.models import Page
 
-from developerportal.templatetags.app_tags import has_at_least_two_filters
+from developerportal.templatetags.app_tags import (
+    filename_cachebreaker_to_querystring,
+    has_at_least_two_filters,
+)
 
 
 class AppTagsTestCase(TestCase):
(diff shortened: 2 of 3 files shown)


developerportal/templatetags/app_tags.py
```diff
@@ -57,3 +61,38 @@ def has_at_least_two_filters(filters):
...
+    404ing even if the cachebreaking hash has changed. (ie, querystring
+    cachebreakers are more forgiving.)
+
+    Note that ManifestFileStorage (used by django-whitenoise in this project) makes
+    hash-named versions of the files, but also _does_ keep the original around,
+    so we are safe to assume that if static/img/foo.213123.jpg exists, then
+    static/img/foo.jpg will, too
+    """
+
+    pattern = re.compile(r"(\.[a-f0-9]+)(\.\w+)$")
```
Gregoor: I'm not the best regex-r so this is definitely more of a question:

Aren't the second set of parantheses superfluous, given that only the the first matching group is used?

stevejalim: True - yes, that second group is (now) unnecessary -- originally i was going the replacement with via `pattern.sub` but it was just getting too complicated. Will remove the second grouping, but keep the search parameters - there are unit test to conifrm if i break things ;)

PR #927, opened 2019-12-03
577: Update footer to link to Mozilla info and Community Guidelines

### LINKED ISSUE #577: Add link to Mozilla mission to footer
From Feedback excel

stevejalim:
The original request (see #577) was to link to Mozilla's mission. Sense-checked where to place this link with Havi, who suggested we actually link to the About page, because that puts the Mission front and centre, in context.

She also suggested we add the Community Participation Guidelines, so I have.

Screenshot 2019-12-03 at 22 20 45

~Note: the existing footer links all have `nofollow` in them -- is this someting we definitely want?~

(Resolves #577)

### DIFF
--- a/developerportal/templates/footer.html
+++ b/developerportal/templates/footer.html
@@ -4,13 +4,15 @@
       <div class="mzp-c-footer-legal">
         <p class="mzp-c-footer-license">
           Portions of this content are &copy;1998–{% now "Y" %} by individual mozilla.org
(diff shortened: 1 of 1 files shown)


atopal:
> Note: the existing footer links all have `nofollow` in them -- is this someting we definitel want?

No, I have no idea why they are nofollow. They shouldn't be.

PR #948, opened 2019-12-05
Update requirements to fix regression when we moved to Wagtail 2.7

stevejalim:
There were too many stale Wagtail-specific dependencies in the lower part of the requirements.txt file, which meant that some things that should have been upgraded for wagtail 2.7 were not. (This was surfaced by Willow 1.1 still being present, and lacking a feature that Wagtail 2.7 expected, causing server errors.)

By removing the "subdeps" from requirements.txt wherever possible, we get a clean install.

The things that HAVE changed now are:
* modelcluster==5.0
* django-taggit==1.2.0
* python-dateutil==2.8.0  # down from 2.8.1 which helps with boto compat
* Willow==1.3

Note that beautifulsoup is pinned to the range that Wagtail needs, because readtime tries to get a newer, incompatible-with-Wagtail version

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ celery==4.3.0
 redis==3.3.11
 kombu==4.6.6  # This is not the latest, but see https://github.com/celery/kombu/issues/1087
 django-celery-results==1.1.2
+django-countries==5.5
 django-redis==4.10.0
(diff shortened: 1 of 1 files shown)


stevejalim:
r+ stevejalim

PR #967, opened 2019-12-12
945: Update pagination and filtering to be done on the server side, not the client side

### LINKED ISSUE #945: Add better pagination and filtering support
We need to move away from the client-side-only show-and-hide logic while also trying to keep as much of the speed while doing server-side side rendering.

If we use querystring args for pagination and filters, the CDN should be able to cache it for us, which will help.

We'll try to avoid the need for a "Apply" button and go with a small timeout before it triggers 

AC
- [ ] `/posts/` page filtering and pagination works and only loads necessary content
- [ ] `/communities/people/` page filtering and pagination works and only loads necessary content
- [ ] `/events/` page pagination works and only loads necessary content
- [ ] Optional? `/events/` page pagination has 'see more' rather than 'see all'
- [ ] Pagination and filtering work in all target browsers, incl mobile
- [ ] State of pagination/filtering is reflected in the URL, so that it's bookmarkable and shareable

stevejalim:
This changeset addresses the problem posed by the current implementation of clientside psuedo paging/filtering: because the page loads all the possible results and then shows/hides DOM elements to match, the overall page weight is already getting significant (eg /posts/ is 10MB and pulls in 18MB of images)

So, when merged, this changeset will switch that to all being handled server-side. The trade-off is we do a full page load each time a page or filter parameter is changed, but if this is too much we can look to update this to AJAX calls + DOM-element replacement.

The original JS for the pagination and filtering has been repurposed and retained wherever possible to still benefit from work done to populate the filter form from querystring params and to show/hide "Clear" triggers appropriately.

## NOTES
- ~The UX of filtering has retained the original behaviour of updating result _immediately_ after ticking a filter box. This now triggers a page reload, which is ok/tolerable on desktop, but more annoying on mobile, because the Filters controls are collapsed when the page is loaded again.~
~This could be dealt with either:~
  I WENT WITH THIS ONE ~1. By changing the filtering to have a 'Filter results' button as the trigger - improving both Mobile and Desktop UX, potentially~
  ~2. On mobile, ensure that if a filter is selected, the Filters panel is kept open on load. This isn't ideal, but would help a little.~
~(My preference is for 1 -- @atopal, what do you think?)~

- DONE ~On Events, future events are not paginated, while past events are. We may well move Past events to their own page, but for now, because past events are so far down the page, viewing the next page is a click-load-scroll down situation.~ 
~This could be dealt with by:~
  * ~Slotting in an #anchor in the pagination links so that the user is auto-scrolled down the appropriate part of the events page after load -- @atopal, worth it?~ DONE

(Resolves #945)

## How to test

TO COME: I'm going to get this branch deployed to a new development environment so that stakeholders can try it out without us having to merge the branch to `master` and we can do direct comparisons in performance and page size with current production. 

If this can't be done soonish, I can also expose by local build over an SSH tunnel - just DM me in Slack and I'll enable it. It won't be as good a comparison, but it'll allow manual test drives.

Note that when there are multiple 'facets' to filtering (eg topic and location) these are combined as `AND` but _within_ each facet, the selections are X `OR` Y.

Areas to test:
- `/posts/`  - try the filtering and pagination. 
- `/events/`  - try the filtering for upcoming events and pagination for past events. Note that the past events listing is not affected by the pagination  
- `/communities/people/` - try filtering and pagination

# Screenshots
Screenshot_2019-12-13 Events
Screenshot_2019-12-13 People
Screenshot_2019-12-13 People
Screenshot_2019-12-13 People(1)
Screenshot_2019-12-13 Posts

Screenshot 2019-12-13 at 12 44 33


atopal:
>       This could be dealt with either:
>       
>       1. By changing the filtering to have a 'Filter results' button as the trigger - improving both Mobile and Desktop UX, potentially
>       2. On mobile, ensure that if a filter is selected, the Filters panel is kept open on load. This isn't ideal, but would help a little.
>          (My preference is for 1 -- @atopal, what do you think?)

Yeah, let's do 1


>     * On Events, future events are not paginated, while past events are. We may well move Past events to their own page, but for now, because past events are so far down the page, viewing the next page is a click-load-scroll down situation.
>       This could be dealt with by:
>       
>       * Slotting in an #anchor in the pagination links so that the user is auto-scrolled down the appropriate part of the events page after load -- @atopal, worth it?

Yeah, but if this turns out not to be straight forward, I'd drop it.

stevejalim:
@atopal thanks for the feedback - will make those small additions you responded to and add them to this PR

Gregoor:
From Slack (didn't know where I'd put it as a code comment, so here it is as a general comment):

The "See more" links on `/topics` still links to the plural `?topics` query param.

stevejalim:
@Gregoor OK, I've made both the changes discussed with Kadir (switching to manual submission triggering + adding an anchor when paging the past events), plus fixed the `topic` vs `topics` bug you spotted (laser eyes!!) -- thanks!

Could you give those aspects a look over, please? ❤️

developerportal/apps/events/models.py
```diff
@@ -132,26 +148,120 @@ def can_create_at(cls, parent):
...
+        past_events = paginate_resources(
+            past_events,
+            page_ref=request.GET.get(PAGINATION_QUERYSTRING_KEY),
+            per_page=self.PAST_EVENTS_PER_PAGE,
+        )
+        return past_events, total_past_events
+
+    def get_relevant_countries(self):
+        # Relevant here means a country that a published Event is or was in
+        raw_countries = set(
```
Gregoor: Can we use [`distinct`](https://docs.djangoproject.com/en/dev/ref/models/querysets/#distinct) here?

developerportal/apps/events/models.py
```diff
@@ -132,26 +148,120 @@ def can_create_at(cls, parent):
...
+        context["show_past_event_pagination"] = (
+            total_past_events > self.PAST_EVENTS_PER_PAGE
+        )
         return context
 
-    @property
-    def events(self):
-        """Return future events in chronological order"""
-        return get_combined_events(self, start_date__gte=datetime.date.today())
+    def _year_months_to_years_and_months_tuples(self, year_months):
```
Gregoor: I think this one deserves a small comment about input/output

developerportal/apps/articles/models.py
```diff
@@ -102,12 +113,27 @@ def can_create_at(cls, parent):
...
 
-    @property
-    def resources(self):
+    def get_resources(self, request):
         # This Page class will show both Articles/Posts and Videos in its listing
-        return get_combined_articles_and_videos(self)
+
+        # We can't use __in in this deeply related query, so we have to make
+        # a custom Q object instead and pass is in as a filter, then deal with
+        # it later
```
Gregoor: I don't understand why `__in` doesn't work here. In Kuma I tested `SocialAccount.objects.filter(user__email__in=['gregorwbr@gmail.com', 'gweber@mozilla.com'])` and it seemed to do the right thing. Does it stop working once the relation is not just between two but three models?

developerportal/apps/events/models.py
```diff
@@ -132,26 +148,120 @@ def can_create_at(cls, parent):
...
+        )
+
+    def get_relevant_dates(self):
+        # Relevant here means a date for a published *future* event
+        # TODO: would be good to cache this for short period of time
+        raw_events = get_combined_events(self, start_date__gte=tz_now().date())
+        return sorted([event.start_date for event in raw_events])
 
     def get_filters(self):
         from ..topics.models import Topic
```
Gregoor: Not your line, but do you know why the local import Circular dependencies? I suppose we should drop a comment about that then

src/js/utils.js
```diff
@@ -13,7 +13,12 @@ exports.parseQueryParams = () => {
   const items = search.slice(1).split('&');
   return items.reduce((acc, item) => {
     const [key, value] = item.split('=');
-    acc[decodeURIComponent(key)] = decodeURIComponent(value).split(',');
+    const val = decodeURIComponent(value).split(',');
+    if (!acc[decodeURIComponent(key)]) {
```
Gregoor: nit: I'd extract that decoded key into its own var

developerportal/apps/articles/models.py
stevejalim: Good question: I tried __in originally, too, but coudn't get it to work, instead triggering a lookup failure. I think there's a limitation going on because of limitations on the `ParentalKey` in the chain of related objects (that non-Django field class comes from `django-modelcluster`). I think the `__` relations can walk more than three in Django, just not when there's a Wagtail `ParentalKey` involved

developerportal/apps/events/models.py
stevejalim: Cheers - updated. Tests stil happy

developerportal/apps/events/models.py
stevejalim: I suspect either original-developer caution, or something that was required before a refactor. Have moved it up to regular imports and things still work, so \o/

src/js/utils.js
stevejalim: Good call - done!

stevejalim:
@Gregoor Thanks for the interim feedback -- all addressed and latest tweaks pushed up. Thanks!

developerportal/apps/articles/models.py
Gregoor: For topics it does seem to work and do the same thing with:
```python
q_object = Q(topics__topic__slug__in=topics) if topics else Q()
```
We might want to investigate switching to this if it works for the other cases as well. What do you think?

stevejalim: @Gregoor Switched to this approach -- thanks for the poke!

Could i get a last look and hopefully a ✅ ?

stevejalim:
Thank _you_ @Gregoor!!

Issue #931, opened 2019-12-04
Move hosting to EKS cluster


stevejalim:
At the moment, the project is running on Kubernetes, but not using AWS EKS, which is @limed's preferred solution for the project. We should move to this as a way to boost stability and uptime of the site (eg we had some short DNS-related DB-availability outages in the past week).

This move is also a prerequisite to moving to dynamic rendering instead of the static site aproach

Acceptance Criteria:

- [x] Staging is running on EKS and generating a static site with zero perceptible changes to end users 
- [x] Production is running on EKS and generating a static site with zero perceptible changes to end users 
- [ ] The old infrastructure is no longer running `developer-portal` code

stevejalim:
@limed Can we tick off the last item on the list in the description and close this?

limed:
Indeed this is all done.

PR #989, opened 2019-12-16
Move celery workers from kops cluster to EKS

limed:
Moves celery workers from kops cluster to EKS

(Related issue #931)

## Key changes:

- Moves hosting from older kops cluster to EKS


limed:
r=limed

PR #1000, opened 2019-12-18
994: Update what counts as a future/past Event 

### LINKED ISSUE #944: Move the Wagtail and Django admins from their defaults

stevejalim:
This changeset adds a single source for the cutoff between past and future events, setting it to 'yesterday'

Prior to this changeset, the difference between future and past events is decided based
on datetime.datetime.now() which is likely to be UTC (given k8s), but regardless of what
specific timezone it is or is not, it still draws a single line "down the globe" that
means that, the further west of UTC you live, some events will 'disappear' from future
events before that day has completed.

The fix is simply to keep showing events as 'upcoming' until one calendar day *after*
their start date.

(Resolves #944)

## How to test

- Add an Event with a start date that is yesterday's date
- On `master`, load the /events/ page and confirm you do not see the relevant Event, even though yesterday's date could still be the current date further westward around the globe
- On this feature branch, load the /events/ page and confirm you no longer see the relevant Event

### DIFF
--- a/developerportal/apps/common/tests/test_utils.py
+++ b/developerportal/apps/common/tests/test_utils.py
@@ -1,0 +1,19 @@
+import datetime
+from unittest import mock
+
+from django.test import TestCase
+
+import pytz
+
+from ..utils import get_past_event_cutoff
+
+
+class EventCutoffTestCase(TestCase):
+    @mock.patch("developerportal.apps.common.utils.tz_now")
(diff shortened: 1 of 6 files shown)


stevejalim:
@valgrimm Looking at the description of this PR, do you think what I'm doing makes sense / is the right call, from a product perspective?

valgrimm:
Yes.

Issue #972, opened 2019-12-13
Move "How to Get Sponsorship for your Developer Event"


valgrimm:
Move out of Communities menu and on to Events page above events list
Steve will add a streamfield above events list and below Featured Event so this can be done
Val will then make the change

AC 
- [ ] Sponsporship link no longer visible in menu
- [x] Events page has streamfield below the featured block that allows adding arbitratry content
- [ ] That streamfield contains a button linking to the sponsorship page

stevejalim:
@valgrimm @atopal I've added AC for this, but in addition, do we want do move where the sponsorship page lives in the CMS (ie, it's under `/community/`). I can add support to allow Events (/events/) to have child pages that are just regular Content Pages too, but that's a separate code change.

Put another way: shall we take the opportunity to allow the Events Page to have children that are Content Pages as well as Event pages? It's probably a couple of hours, incl QA

valgrimm:
Hi Steve, thank you for this and the well-described proposed extension. 
I'm not sure in what other context we'll need non-Event child pages for the events area, so I propose skipping that extra functionality for now. (I'd like that for Topics, but that's another ticket and again not sure that is needed either). 

So short version, just move where the link appears, not where the content lives. 
Thanks!

atopal:
+1 agreed, and yes, thanks for adding the AC!

valgrimm:
Steve enabled adding the button, and I have added it to the events page to link to  "How to Get Sponsorship for your Developer Event", and removed  "How to Get Sponsorship for your Developer Event" from the Communities dropdown.

PR #1003, opened 2019-12-18
972 Add streamfield to events page to support custom non-Event content

stevejalim:
This changeset addresses the need to link to the sponsorship page _from_ the /events/ page by adding support for free-form rich content (as per our custom StreamField's config)

(Related issue #972)

## How to test

- Edit the Events page - add some rich content for the Body element and publish or preview the page. Depending on what you add, you'll see something like this:

Screenshot 2019-12-18 at 13 50 20


stevejalim:
@valgrimm You can now test this out on the Dev env. Let me know if it does all you need it to and i'll get it shipped to production

valgrimm:
This looks great. 

(In fact, it gave me an idea for some other changes I'd like to make elsewhere on the site, but I first proposed them to Kadir before creating any tickets or asking about dev (only one requires dev). Thanks!)

Issue #872, opened 2019-11-26
Provide way to append comments to "Submit for Moderation" requests


valgrimm:
For example: "Release on December 10th at 10am PST" or "I added a subject tag to my profile"

stevejalim:
That sounds like a nice feature to add to Wagtail itself. 

If we can allocate the time, I'd be happy to explore adding that as a contribution back to Wagtail

valgrimm:
@steve: I upvote. Tell Matt it was me (:

atopal:
Steve will see if this can be added to wagtail

stevejalim:
We can write up and submit an RFC at https://github.com/wagtail/rfcs

valgrimm:
Super. I shall write something up and send it to you for input and we can meet. Will do during next sprint if ok by Kadir. have assigned to me so I write first draft of req.

atopal:
yeah, works for me!

valgrimm:
I have created an initial draft of the RFC minus technical detail. Now I am adding it to this ticket and assigning the ticket to Steve so that where there is time he can review what I have written add this detail.
wagtail-rfcs/000-append-comments-for-moderation.md

I read the instructions for writing Wagtail RFCs before writing this RFC (https://github.com/wagtail/rfcs).

PR #1055, opened 2019-12-30
872: Tweak and expand the RFC for content moderation messages

stevejalim:
Added initial technical specification section and tweaked a bit of the great initial spec

Some of the changes here are just down to auto-formatting from the code editor.

If/once this is ok we can merge it to `master` then submit it via pull request from a fork of `wagtail/wagtail`, as per the official guidance

(Related issue #872)

### DIFF
--- a/wagtail-rfcs/000-append-comments-for-moderation.md
+++ b/wagtail-rfcs/000-append-comments-for-moderation.md
@@ -1,53 +1,92 @@
 # RFC 1: Provide option to require a comment when submitting content for moderation
 
-* RFC: 1
-* Author: Steve Jalim and Val Grimm
-* Created: 2019-12-21
-* Last Modified: 2019-12-23
+- RFC: 1
+- Authors: Val Grimm (@valgrimm) and Steve Jalim (@stevejalim)
+- Created: 2019-12-21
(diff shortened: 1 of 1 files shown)


wagtail-rfcs/000-append-comments-for-moderation.md
```diff
@@ -1,53 +1,92 @@
...
+- This functionality will not guarantee an improved workflow: Editors can still circumvent it by typing nonsense into the comment field.
 
-### Drawbacks
-* A developer will be needed to deactivate the functionality as described above.
-* In some situations the functionality will not be needed.
-* When this functionality is implemented, it will be easy for editors to circumvent by typing nonsense into the comment field. 
+## Alternatives
 
-### Alternatives
-* More detailed linting functionality to enable administrators to ensure that comment submissions are not spurious.
```
stevejalim: @valgrimm I dropped this line because I noticed that Wagtail does have a useful way to see what has changed in a page, even if it it doesn't cover the _why_ of the change.

I wondered whether this would be useful (or even - at a long shot - enough)? 


Screenshot 2019-12-30 at 13 27 14

PR #1056, opened 2019-12-30
1011: Fix premature nav reflow issue 

### LINKED ISSUE #1011: Nav reflow/overflow - needs addressing?
The change to "Products and Technologies" and the addition of "About" in the Nav means it reflows and gives us a lot of whitespace in the main nav and kind of takes us off-brand (compare to https://protocol.mozilla.org/demos/navigation.html)

If we want to remove that whitespace, we could

* move "About" down in to the footer
* rephase "MDN Web Docs" as just "MDN" (if that's not against brand rules)
* rephrase "Products and Technologies" to just "Technologies" (if that's not going against stakeholder requests)

What do you both think @valgrimm @atopal ?

stevejalim:
This changeset fixes the reflow that occurs when the nav options are particularly numerous and/or long.

Basically, it makes the CSS for the largest viewport use the same, slightly reduced padding values as the next, smaller media query.

_Note that things **are** quite packed in now. If we can move About to the footer, regardless of this change, that would still be useful_

(Resolves #1011)

## How to test

Take a look at https://developer-portal.dev.mdn.mozit.cloud/, which originally was like:

Screenshot 2019-12-30 at 21 40 49

And is now

Screenshot 2019-12-30 at 21 55 45

*Please have a play around with the viewport sizes by resizing your browser and trying to break the layout* 😄

### DIFF
--- a/src/css/molecules/header.scss
+++ b/src/css/molecules/header.scss
@@ -71,6 +71,12 @@
     cursor: pointer;
     font-weight: normal;
 
+    @media #{$mq-xl} {
+      // OVERRIDE to make it the same as $mq-lg, so we can pack more in to the nav
+      // without awkward reflow
+      padding: 0 16px 24px;
+    }
+
     img {
       margin-right: 8px;
     }


valgrimm:
This looks good, I'm just seeing one weird behavior when shrinking the viewport. At the narrowness setting just before the top nav turns into a hamburger, the nav line breaks just before Firefox. If this needs to be fixed, then fix it. But otherwise looks fine to me. 

weirdbreak1011

PR #1079, opened 2020-01-08
Update RBAC for jenkins

limed:
Updated RBAC for jenkins user to now allow creation of HPA's

## Key changes:

- Updated RBAC to allow creation of HPA


limed:
r=limed

Issue #1017, opened 2019-12-20
Implement automated blog+YouTube syndication


valgrimm:
Ideally, syndicating developer-relevant blogs from around Mozilla on the portal is part of the goals for 2020. 
A content task to identify relevant blogs was also created (see #1016). 

Revised AC: 
At an interval of two hours for YouTube and 30min for blogs (can correct timing if unrealistic).
- [ ] Entry-level: Posts from relevant blogs+YouTube channels (manually added to a list in code somewhere) are automatically added as "External Posts" on the portal. 
- [ ] Mid-level: Social sharing image associated with each post is automatically filled into image field so the placeholder image doesn't have to be replaced.
- [ ] Fancy-unrealistic-level: A tag is automatically assigned depending on the blog/YouTube.

(Perhaps I should create an Epic to link these tickets).

stevejalim:
Quick estimates, assuming all blogs have an RSS feed (plus CR and acceptance time):
entry level: 1.5 days
mid level: 2 days, inclusive
fancy: 2.5 days, inclusive

Youtube: +0.5 day, as long as RSS feed remains alive: https://www.youtube.com/feeds/videos.xml?channel_id=UCh5UlGiu9d6LegIeUCW4N1w

valgrimm:
Inclusive!
And add YouTube to AC. 
I shall edit ticket title accordingly.

valgrimm:
Now unblocked: I found all the blogs and youtube channels we will have for the near future

stevejalim:
@valgrimm and @havi Quick question: for the automatic post/video ingestion, do you want the things to be automatically published, or simply submitted for moderation?

I'm leaning towards the latter, because it gives us better visibility of what's going up - including a chance to skip any items that we might not want linked to from the portal (for some hand-wavy reason I can't think of right now :smile: )

valgrimm:
I defer this to @havi but generally agree with Steve. Kudos on thinking of this Steve!

stevejalim:
@valgrimm Got another couple of questions: we can pull in data about authors, which I can then try to match with existing `Person` pages we hold. 
1) Do you think this is worth it, when we can clearly match it to a single `Person`? (If there are multiple matches, somehow, I think the only safe move is to not auto-associate a Person with the newly created Page.

[...]

PR #1100, opened 2020-01-14
1017: Automatic content ingestion

stevejalim:
This changeset introduces support for automatically adding draft content to the CMS and pinging moderators to approve it.

It pulls data from RSS or Atom feeds and comes with a data migration that automatically configures an initial set of Blogs and YouTube feeds to pull from:

Screenshot 2020-01-14 at 14 09 51

The background scheduler runs two tasks every two hours (but not at the same time - 30 mins apart), which will attempt to ingest videos or articles respectively. If the relevant feeds show that there's a new item **created** since the last time the feed was checked, it is added to the CMS as draft content and the Moderators are emailed to notify them about this:

```
-------------------------------------------------------------------------------
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: The page "Mozilla Announces Deal to Bring Firefox Reality to Pico
 Devices" has
 been submitted for moderation
From: None
To: notsjalim@mozilla.com
Date: Mon, 13 Jan 2020 11:19:35 -0000
Message-ID: <157891437597.201.13313017575110632161@de69f79bfd63>
Auto-Submitted: auto-generated

Hello Editor,



The page "Mozilla Announces Deal to Bring Firefox Reality to Pico Devices" has been submitted for moderation by Portal Bot.

You can preview the page here: http://localhost:8000/admin/pages/moderation/359/preview/
You can edit the page here: http://localhost:8000/admin/pages/151/edit/



Edit your notification preferences here: http://localhost:8000/admin/account/notification_preferences/
```

The Moderator can then preview, accept or reject the item. Note that rejecting it will still retain the draft version, so it's not a terminal action.

Screenshot 2020-01-13 at 12 34 14

Note that because we're pulling in data from various sources, we need to be sure that the slugs of the items are still unique even if two things have the same title, so we append a hash of the data payload extracted from the feed (this payload includes contains the title, authors and a timestamp, so is fairly reliable for resulting in a unique hash). eg `/videos/we-made-another-channel-35017c1cfa13`

(Related issue #1017)

## How to test

* Ensure your local stack is configured for AWS so that it can save Images to S3 (or that you're sure it can save Images to local disk instead of needing S3)
* When booting up, confirm the scheduled ingestion tasks are registered to run two-hourly

```
scheduler_1  | <ScheduleEntry: run-publish-scheduled-command-every-hour developerportal.apps.taskqueue.tasks.publish_scheduled_pages() <crontab: 55 * * * * (m/h/d/dM/MY)>
scheduler_1  | <ScheduleEntry: selectively-purge-cdn-every-night developerportal.apps.taskqueue.tasks.selectively_invalidate_cdn() <crontab: 15 0 * * * (m/h/d/dM/MY)>
scheduler_1  | <ScheduleEntry: ingest-videos-every-two-hours developerportal.apps.ingestion.tasks.ingest_videos() <crontab: 37 */2 * * * (m/h/d/dM/MY)>
scheduler_1  | <ScheduleEntry: ingest-articles-every-two-hours developerportal.apps.ingestion.tasks.ingest_articles() <crontab: 17 */2 * * * (m/h/d/dM/MY)>
scheduler_1  | <ScheduleEntry: celery.backend_cleanup celery.backend_cleanup() <crontab: 0 4 * * * (m/h/d/dM/MY)>
scheduler_1  | [2020-01-14 13:31:55,252: DEBUG/MainProcess] beat: Ticking with max interval->5.00 minutes
```

and that the tasks to ingest articles and videos are also registered at the worker level.

```
...
worker_1     |   . developerportal.apps.ingestion.tasks.ingest_articles
worker_1     |   . developerportal.apps.ingestion.tasks.ingest_videos
...
```

* Set all IngestionConfiguration `last_sync` dates to 2019-12-12
* `docker-compose exec app python manage.py shell`
* `from developerportal.apps.ingestion.tasks import *`
* Execute `ingest_videos()` - you should see "INFO:developerportal.apps.ingestion.tasks:Skipping automatic video ingestion because settings.AUTOMATICALLY_INGEST_CONTENT is False"
* Edit your `local.py` to set `AUTOMATICALLY_INGEST_CONTENT = True`
* Quit the Django shell and re-enter it.
* Execute `ingest_videos()` again - you should see some logging of it happening, as well as emails going to the console saying that new items are ready for moderation (see below for an example of that)
* Flip to locahost:8000/admin/ - you should see a moderation queue
* Flip to localhost:8000/posts/ - confirm nothing in the moderation queue appears yet in the live page
* In the admin, preview a draft/pending Video page (opens a new tab), then close the tab
* In the admin, back on the moderation list, approve that Video
* Confirm it now appears in localhost:8000/posts/ complete with a thumbnail
* Reject a video and confirm it does not appear in the /posts/ page after that (but the Draft page does still exist in Wagtail)

* Now do the same for `ingest_articles()`, which creates ExternalArticles (ie, links to externally hosted content, rather than creating pages inside the CMS that hold the same content) 
* Confirm that previewing them opens a new tab, outside of the CMS;
* approve some, reject others, as above, confirming presence/absence from /posts/ accordingly.
* Re-run the two `ingest_*()` functions and note that duplicates are gracefully handled:
```WARNING:developerportal.apps.ingestion.utils:Problem ingesting article from {'title': 'Mozilla Hacks’ 10 most-read posts of 2019', 'authors': ['Havi Hoffman'], 'url': 'https://hacks.mozilla.org/2019/12/mozilla-hacks-most-read-blog-posts-of-2019/', 'image_url': '', 'timestamp': datetime.datetime(2019, 12, 17, 17, 22, 57, tzinfo=tzlocal()), 'owner': <User: ingestion_user>}: {'slug': ['This slug is already in use']}```

Note that this code is currently running on the dev env at https://developer-portal.dev.mdn.mozit.cloud/, so we'll see how the automatic ingestion works on a schedule

### CODE REVIEW

**Comment by stevejalim** [comment]
@valgrimm Please could you keep an eye out for notification emails from the dev site - if you get one, feel free to follow the link and try out moderating an auto-generated item of content.

File: developerportal/apps/ingestion/tests/data/test_images_as_bytearrays.py [hunk 1]
```diff
@@ -0,0 +1,3 @@
+image_one = bytearray(
```
stevejalim: You can ignore this in code review

k8s/Makefile
```diff
@@ -35,6 +35,7 @@ export APP_EMAIL_HOST ?= email-smtp.us-west-2.amazonaws.com
 export APP_EMAIL_PORT ?= 587
 export APP_EMAIL_USE_TLS ?= True
 export APP_SECURE_HSTS_SECONDS = 31536000
+export APP_AUTOMATICALLY_INGEST_CONTENT ?= True
```
stevejalim: Heya @limed could you confirm i have all the pieces of the puzzle to get `settings.AUTOMATICALLY_INGEST_CONTENT` set up as `True` in all environments, please?

valgrimm:
> 
> 
> @valgrimm Please could you keep an eye out for notification emails from the dev site - if you get one, feel free to follow the link and try out moderating an auto-generated item of content.

Hi Steve, I got the notification emails for the items added by the bot and was able to look at and publish them. Thanks!

developerportal/apps/ingestion/tests/test_data/blog.mozvr.com_rss.xml
```diff
@@ -0,0 +1,506 @@
...our device back to Facebook servers during a routine browsing session? From the data we saw, they’re reporting configuration and telemetry data, such as information about how long it took to fetch resources. For example, here’s a graph of the amount of data sent over time from the Oculus VR headset back to Facebook.</p><!--kg-card-begin: image--><figure class="kg-card kg-image-card kg-card-hascaption"><img src="https://blog.mozvr.com/content/images/2019/12/Screenshot-2019-12-03-14.06.51-1.png" class="kg-image" alt="How much is that new VR headset really sharing about you?"><figcaption>Bytes sent to Facebook IPs over time</figcaption></figure><!--kg-card-end: image--><p>The data is identified by both an <code>id</code>, which is consistent across browsing sessions, and a <code>session_id</code>. The <code>id</code> appears to be linked to the device hardware, because linking a Facebook account didn’t change the identifier (or any other information as far as we detected).</p><p>In addition to general timing information, Facebook also receives reports on more granular, URL level timing information that uses a unique URL ID.</p><!--kg-card-begin: code--><pre><code>"time_to_fetch": "1",
```
stevejalim: This can be ignored - it's a saved version of a Mozilla VR RSS feed from new year

developerportal/apps/ingestion/tests/test_data/youtube_feed.xml
```diff
@@ -0,0 +1,463 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
stevejalim: This can be ignored - it's a saved version of a YouTube Atom feed from new year

k8s/Makefile
limed: You will need to expose this in `app.env.yaml.j2` file as well

limed: > You will need to expose this in `app.env.yaml.j2` file as well

Didn't scroll down enough, looks like you did do that so you should be good :D

developerportal/apps/ingestion/utils.py
```diff
@@ -0,0 +1,261 @@
...
+                        notification="submitted",
+                        excluded_user_id=ingestion_user.id,
+                    ):
+                        logger.warning(
+                            "Failed to send notification that %s was created.",
+                            draft_page,
+                        )
+
+
+def _store_external_image(image_url: str) -> MozImage:
```
Gregoor: shouldn't this also be `@transaction.atomic()`?

developerportal/apps/ingestion/utils.py
```diff
@@ -0,0 +1,261 @@
...
+    model_name = type_
+
+    configs = IngestionConfiguration.objects.filter(integration_type=type_)
+
+    Model = _get_model(model_name)
+
+    ingestion_user = User.objects.get(username=INGESTION_USER_USERNAME)
+
+    for config in configs:
+        with transaction.atomic():
```
Gregoor: I wonder if we'd want to have a mechanism in this function to ensure it's not running multiple times for the same types at the same time. But before I indulge those thoughts: what would happen in that case? Would the second transaction just error out because of some constraints or would we have duplicated items?

developerportal/apps/ingestion/utils.py
```diff
@@ -0,0 +1,261 @@
...
+                except ValidationError as ve:
+                    logger.warning("Problem ingesting article from %s: %s", data, ve)
+                else:
+                    draft_page.owner = ingestion_user
+                    draft_page.save()
+                    # This is a little messy: we save a new revision and ping the
+                    # moderator, but because it's within a transaction if something
+                    # fails we will roll it back at the DB level, but
+                    # the emails notifying Moderators will have been sent for pages
+                    # which will not exist. It's not ideal, but safer than having
```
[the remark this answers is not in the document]
stevejalim: Looking back at this now, I think we could build a 'success buffer' containing new pages/items and then trigger the notification for each after the transaction is commited. 

Does that feel better to you?

developerportal/apps/ingestion/utils.py
```diff
@@ -0,0 +1,261 @@
...
+def generate_draft_from_external_data(model, data, **kwargs):
+    """Create a draft page of the appropriate `model` (eg: ExternalArticle,
+    Video) from the given `data`, including any associated thumbnail
+    (which is saved down as a Wagtail image).
+
+    It uses a SHA1 hash of all the data for the slug, because it'll be unique
+    (because it contains a datetime) and deterministic -- and doesn't get
+    shown to the users anyway.
+    """
+    if model not in (externalcontent_models.ExternalArticle, video_models.Video):
```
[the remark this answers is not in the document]
stevejalim: Ah yeah, that's a great point. This func got moved over from a model method in a separate module, and I missed this duplication during the refactor. This block should never be triggered, in real terms, so we can drop it. Thanks for querying it!

developerportal/apps/ingestion/utils.py
stevejalim: I didn't put it in a transaction because it only touches one model/one DB table, so there's no related-but-not-failed state to roll back. It's called from something that _is_ inside an atomic transaction, though, so if this or something else in that transaction fails, they all get rolled back at the DB level (if not at the file-storage level – which could be nice to do, but not essential I think)

developerportal/apps/ingestion/utils.py
stevejalim: Ah yeah, we do have a Redis-backed lock we could use (we used to use it to avoid two static builds happening at the same time and fighting with each other). I nearly put it in here, but I thought for the initial version it wasn't worth it - if we end up with potential duplicates, these raise ValidationErrors at save() and are gracefully logged and we move on. 

The only thing I'm not 1000% sure on is if _both_ of two identical imports are happening at te same time, and they're both within their own transaction, whether duplicates will be detected because the commit will happen, in theory, after Python has done checking for a ValidationError based on the slugs committed to the DB at the point the transaction started, if you see what I mean. What might then happen is an IntegrityError, because slugs must be unique - and we'll hear that loudly in Sentry.

developerportal/apps/ingestion/utils.py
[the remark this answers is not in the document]
Gregoor: That does sound better to me. Or maybe even batched together into one email? Btw where is the email template?

developerportal/apps/ingestion/utils.py
Gregoor: That does make sense and alleviates my fears. Thanks for the elaboration!

developerportal/apps/ingestion/utils.py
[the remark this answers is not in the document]
stevejalim: The email template is [from Wagtail](https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/wagtail/admin/templates/wagtailadmin/notifications/base.txt) but we could override it if we want to

Gregoor:
I'm starting to think we should do some dependency inversion (if this is what I think it is) in `generate_draft_from_external_data` and instead of passing a model into it, we pass a function into it that initializes the correct page model, which should turn this
https://github.com/mdn/developer-portal/blob/248a005b423ed6706ea0599405760ab137fdf5a7/developerportal/apps/ingestion/utils.py#L227-L252
into something like
```python
def generate_draft_from_external_data(make_page, data, **kwargs):
  ...
  parent_page, page = make_page(data.get("url"))
  parent_page.add_child(instance=page)
```
and then all the model class specific branching stuff can be defined outside of the general flow of this fn (which imo helps understanding).

Am I making sense?

---

Apart from that it looks and runs fine for me and I was not able to shoot holes into it locally :)

developerportal/apps/ingestion/utils.py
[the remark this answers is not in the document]
stevejalim: Sticking with one-email-per-item for now, because it involves replumbing a bit deeper than I have headspace for right now

developerportal/apps/ingestion/utils.py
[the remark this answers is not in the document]
stevejalim: @Gregoor addressed in 48aeb600

developerportal/apps/ingestion/utils.py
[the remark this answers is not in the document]
stevejalim: @Gregoor See 17326f81e0 - refactored (but for now still sending one email per item)

developerportal/apps/ingestion/utils.py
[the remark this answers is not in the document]
stevejalim: Aaaand `_get_models` was refactored/renamed/repurposed in b1588edc4, so the specifics of 48aeb60 are less relevant now, other than the idea that they remove duplicated effort.

PR #1110, opened 2020-01-16
941: Remove redundant static-build code/behaviour

### LINKED ISSUE #941: Remove the static-build pipeline from the codebase
AC

- [x] Remove the Django code that triggers the static build process when a page is published
- [x] Remove the code that runs a scheduled task to try to build a static site every minute.
- [x] Update the project dependencies to no longer use the wagtail-bakery package that generates the static site 
- [x] Remove tests for code related to static pipeline 
- [x] Update developer documentation to no longer talk about the static site

stevejalim:
This changeset removes the ability to build (automatically or on demand) a static site and publish it to S3.  We've already moved away from using the S3-hosted site for production when we introduced traditional server-side pagination, so this is a cleanup ticket. No _active_ functional changes are included here, only dropping now-dead code

(Resolves #941)

## Key changes:

- Remove `staticbuild` app from the codebase, which contained all the async tasks and hooks to make automatic build-and-sync run
- Remove `wagtail-bakery` and our custom views for it, which handled the actual building and publishing of the static site
- Add a schema migration to drop the `staticbuild_staticbuild` DB table left behind, orphaned when the `staticbuild` app was removed

## How to test
The code is up on https://developer-portal.dev.mdn.mozit.cloud - have general play around in the CMS and confirm you can publish changes to the content fine. (Because this is a removal PR, checking for things working as expected is all we can do)


valgrimm:
I tested in the following way:
Create an:
Event
Post
Video
External post
and Topic.
I could publish all of these except Topic.
With Topic I tried several ways and receive the error "The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application."
There may be a reason for this that I did not think of, so best someone else checks Topic.
I can approve once Topic is proven ok.

stevejalim:
@valgrimm 

Thanks for the test drive!

> I could publish all of these except Topic.

The reason Topic was failing is - thankfully - unrelated to these changes. (The state of the database on dev still matches the changs the 'hide topics from list views' feature, but the code for that feature was taken off dev.)

I've pushed an updated branch (with the topics-related changes in there too, to keep the DB happy. If you test again now, it should be fine!

PR #1112, opened 2020-01-16
729: Enable RSS feed for content that lives on the portal

### LINKED ISSUE #729: Re-enable RSS feed
We've got support an RSS feed served by the CMS, but we were not exporting it to the static site, so the link was 404ing. As part of #625 it was disabled temporarily, but once we've moved to the dynamic site (PR #1001) we can re-enable it

AC
- [ ] RSS Feed for posts is available from the CMS site

stevejalim:
This changeset re-enables the RSS feed we previously had created but parked (because it wasn't compatible by default with the static-site approach).

We now render a node in the <head> for the RSS feed for Posts content only:

      <link rel="alternate" type="application/rss+xml" href="/posts-feed/">

In re-enabling it, I've bumped the max number of entries to 40, from 20.

Also note that this feed ONLY shows content from /posts/ that lives on the portal, not any external links to posts/videos. As such, with the automatic content ingestion that's due to land soon, we may find that most of the content in the RSS feed is for videos (which trigger the creation of Video pages in the CMS), because links to blog posts, etc, are rendered as external links (so don't appear in this feed)

(Resolves #729)

## How to test

- view source on any page on http://developer-portal.dev.mdn.mozit.cloud/ and you should see the link to  /posts-feed/ in the <head>
- follow that link and see the RSS XML content that links off to relevant posts - none of these will be ExternalArticles, they will all be Articles/Posts or Videos


stevejalim:
@valgrimm Can I get a behaviour check on this PR from you (as per the How To Test bit in the description) please? Thanks!

valgrimm:
Works now, thanks!

stevejalim:
Not adding this any more/yet. Will put the branch in the icebox for now

Issue #1010, opened 2019-12-19
Change or remove text header of Topics panel


valgrimm:
Right now the text header in the Topics panel reads:
"Browse by topic"

Could this be easily changed to "Browse Products and Technologies" or removed?

AC / areas to fix:
- [x] "Browse by topic" on homepage
- [x] "Browse by topic" on topics list
- [ ] Video page "Read more on this topic"
- [ ] Post page "Read more on this topic"
- [x] Posts filters "Topic"
- [x] Events filters "Topic"
- [x] People filters "Topic"
- [x] Person page "Topics I specialize in"

stevejalim:
@valgrimm Yep, can def be changed. What's your preference: rename or remove?

valgrimm:
Rename is definitely my preference.

stevejalim:
I'll see if i can get it to pull in whatever the _title_ of the page currently known as "Topics"

valgrimm:
works for me

stevejalim:
@valgrimm This change to "using whatever the Topics page is called" is now live, so if you rename Topics to Products & Technologies, you will see the phrasing update elsewhere in the site, too

valgrimm:
Fixed. Thanks! (:

PR #1119, opened 2020-01-20
1010 Make "Topics" phrasing dynamic

stevejalim:
This changeset addresses the need to stop calling [most] things "Topics" and instead draws the phrasing to use from the current title of the `Topics` page. It does this in a way that minimises database hits just to get that display string.

(Related issue #1010)

## Key changes:

- Update base/common fixture with new phrasing of Topics page title, for local consistency
- Add Python-layer support for dynamically using the title of the Topics page in rendered output:
  * adding a context processor to make the Topics title string available in all templates
  * added write-through caching behaviour to the above, so we don't hit the DB unnecessarily
  * added cache invalidation via a 'bulk invalidate' approach: when a Topics model is saved, all its related/configured
cache keys are zapped
  * tests for the above

- Use the new context variable for the title of the topics page in HTML

## How to test

- have a look around on the [dev](https://developer-portal.dev.mdn.mozit.cloud/) server, but also here are some screenshots to show what has/has not changed:

**Homepage and `/topics/`**
Screenshot 2020-01-20 at 15 42 29

**Person page when they are associated wtih one or more `Topic`s**
Screenshot 2020-01-20 at 15 43 27

**Filtering sidebar**
Screenshot 2020-01-20 at 15 46 53
Screenshot 2020-01-20 at 15 47 06
Screenshot 2020-01-20 at 15 47 12

**Note that this one has NOT changed (sidebar on Post and Video pages) because I couldn't find a neat way to reuse the title text available from the Topics page**
Screenshot 2020-01-20 at 17 41 34

### DIFF
--- a/developerportal/apps/common/models.py
+++ b/developerportal/apps/common/models.py
@@ -1,3 +1,5 @@
+from django.core.cache import cache
+from django.db import transaction
 from django.db.models import SET_NULL, ForeignKey
 
 from wagtail.core.models import Page, PageManager
@@ -26,5 +28,14 @@ class BasePage(Page):
         related_name="+",
     )
 
+    #  A single place to store/define cache keys that need invalidation
+    # when the object is saved
+    _bulk_invalidation_cache_keys = []
+
     class Meta:
         abstract = True
+
+    @transaction.atomic()
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        cache.delete_many(self._bulk_invalidation_cache_keys)
--- a/developerportal/context_processors.py
+++ b/developerportal/context_processors.py
@@ -1,4 +1,9 @@
+import logging
+
 from django.conf import settings
+from django.core.cache import cache
+
+logger = logging.getLogger(__name__)
 
 
 def google_analytics(request):
@@ -28,6 +33,29 @@ def directory_pages(request):
             "topics": Topics.published_objects.first(),
         }
     }
+
+
+def topics_title(request):
+    """Get a name to use for the section formerly known as "Topics", based on
+    whatever the Topics page is currently called.
+
+    NB: this assumes the Topics page's title is a plural."""
+    from .apps.topics.models import Topics
+
+    _key = Topics.CACHE_KEY_TOPICS_TITLE
+
+    title = cache.get(_key)
+    if not title:
+        topics_page = Topics.published_objects.first()
+        if not topics_page:
+            # Something is very wrong here, but temporarily fall back regardless
+            # without caching
+            return {"TOPICS_TITLE_LABEL": "Topics"}
+
+        title = topics_page.title
+        cache.set(_key, title, settings.CACHE_TIME_VERY_LONG)
+
+    return {"TOPICS_TITLE_LABEL": title}
 
 
 def pagination_constants(request):
--- a/developerportal/apps/common/fixtures/common.json
+++ b/developerportal/apps/common/fixtures/common.json
@@ -654,8 +654,8 @@
       "path": "000100010002",
       "depth": 3,
       "numchild": 2,
-      "title": "Topics",
-      "draft_title": "Topics",
+      "title": "Products & Technologies",
+      "draft_title": "Products & Technologies",
       "slug": "topics",
       "content_type": [
         "topics",
--- a/developerportal/apps/common/tests/test_context_processors.py
+++ b/developerportal/apps/common/tests/test_context_processors.py
@@ -1,0 +1,40 @@
+from unittest import mock
+
+from django.core.cache import cache
+from django.test import TestCase
+
+from developerportal import context_processors
+
+from ...topics.models import Topics
+
+
+class ContextProcessorsTestCase(TestCase):
+
+    fixtures = ["common.json"]
+
+    def setUp(self):
+        cache.clear()
+
+    def tearDown(self):
+        cache.clear()
+
+    def test_topics_title(self):
+
+        self.assertEqual(Topics.objects.get().title, "Products & Technologies")
+        self.assertIsNone(cache.get(Topics.CACHE_KEY_TOPICS_TITLE))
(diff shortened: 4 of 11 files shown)


developerportal/apps/common/models.py
```diff
@@ -26,5 +28,14 @@ class BasePage(Page):
         related_name="+",
     )
 
+    #  A single place to store/define cache keys that need invalidation
+    # when the object is saved
```
stevejalim: We could do this differently and use redis more directly as a cache, in order to get wildcard invalidation or tagged sets of keys to invalidate

developerportal/context_processors.py
```diff
@@ -30,6 +35,29 @@ def directory_pages(request):
...
+    whatever the Topics page is currently called.
+
+    NB: this assumes the Topics page's title is a plural."""
+    from .apps.topics.models import Topics
+
+    _key = Topics.CACHE_KEY_TOPICS_TITLE
+
+    title = cache.get(_key)
+    if not title:
+        topics_page = Topics.published_objects.first()
```
Gregoor: Wondering if dependening on the published order is not a little brittle. Could we maybe use something like the page id to retrieve the title? I guess right now we're only caching it for a single page, so maybe I'm jumping the gun.

stevejalim: In a non-Wagtail world, I totally agree with you, but the structure of the site is such that there will (in real terms) only ever be one Topics page -- if there is more than one, the site will be effectively broken/in a weird state. As such, expecting just one `Topics` page to exist is not an unreasonable expectation.

Does that help ease your (excellent) conscience?

developerportal/apps/common/models.py
```diff
@@ -26,5 +28,14 @@ class BasePage(Page):
...
+    _bulk_invalidation_cache_keys = []
+
     class Meta:
         abstract = True
+
+    @transaction.atomic()
```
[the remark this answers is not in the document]
stevejalim: Ah yeah, I shoulld have made this clearer: yes, this PR is partly laying in a PoC/approach for adding more cacheing and cache-invalidation to the other pages in the site too.

If you think there's a nicer way, in general, re cache handling, I'm definitely open to it.

developerportal/apps/common/models.py
```diff
@@ -26,5 +28,14 @@ class BasePage(Page):
...
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        cache.delete_many(self._bulk_invalidation_cache_keys)
```
Gregoor: how about we only delete the cache key for the currently saved record, i.e. use `self.pk` in the cache key?

developerportal/context_processors.py
Gregoor: Shall we then maybe use the `pk` defined in the fixtures instead of relying on the order of topics? That would feel better to me. I think I'm feeling better about making the general assumption about there only being one topics page. Thanks for the elaboration :)

stevejalim: > Could we maybe use something like the page id to retrieve the title

Given the autoincrement state of the DB may be different across local dev or the dev|stage|prod k8s envs, `.first()` seems less brittle than going with IDs. And because the `/topics/` slug isn't guaranteed either, I think what we have is prob the best pick of the bunch

developerportal/apps/common/models.py
stevejalim: Hmm - maybe, yeah. I think the next piece of work re cacheing (which will broaden out how we do it across models and will inevitably have some tuning involved) might be a better place to explore that. Right now, `Topics._bulk_invalidation_cache_keys` is a list with just one member, so there's not much to gain in efficiency/specificity terms.

Could you live with this as-is for now?

I quite like the idea of using `self.pk` to as a label/tag for a set of keys to do bulk invalidation on (via something like https://gist.github.com/stevejalim/a55d53994a455dd3e95e)

developerportal/context_processors.py
Gregoor: Oh, so the pk in the fixtures
https://github.com/mdn/developer-portal/blob/34174f1313287d16f6d5f5fd7f79e853ec0f043c/developerportal/apps/common/fixtures/common.json#L652-L657
is not necessarily respected?

Huh tough nut indeed!

stevejalim: Thinking about it: in tests and local setup, yes, it's used, but not in all the k8s deployments, where pages in Prod and Dev were created by hand. Stage may have used common.json tho

Gregoor: I see, dang! Doesn't that then open even more room for them to be mis-ordered? Could we maybe check (once) which page is the one that contains all the topics and then use that one here?

developerportal/apps/common/models.py
Gregoor: Fair! :+1: I'm still figuring out how to balance what's currently needed from the caching thing with what will be needed which I probably don't know enough about. So overall I do think it's fine for now and I hope I still remember all those thoughts when I review the next cache PR.

developerportal/context_processors.py
stevejalim: Given that we "trust" that there will only be one `Topics` page (which is a unique subclass of `BasePage` and the only page that can have `Topic` (singular) pages as children), I don't think we'll have a problem with the wrong `Topics` page being picked. 

However, your point does make me think that we can add a constraint to certain page types (ie, `Topics`, `Articles`, `Events`, `People`) to enforce the idea that we can only allow one single instance of these structural/grouping pages to exist. That would solve this risk of multiple `Topics` pages, by protecting us from actually adding more than one. Does that work for you?

Gregoor: Oh, I think I was confused about that. There'll only ever be one instance of `Topics`? I guess everything I would ask as a follow-up would break the scope of this PR and would fall more under general database design curiosity. So for this thing here I'm in agreement, thanks for enlightening me :)

Issue #1132, opened 2020-01-24
Check and monitor odd behaviour with auto ingestion


stevejalim:
1) stage shouldn't have sent a notification message, but it did
2) where's the new content on prod that matches the content that was notified by on stage

stevejalim:
Seems stable - just ingested two more things and notified from both prod and stage, as expected

stevejalim:
Hmm just had it work fine on prod but not invest on stage despite both trying at the same time

Nothing in the monitoring logs

stevejalim:
Leaving this open to gather more info about any unusual behaviour. Ideally we'll get the same messages from stage and from prod announcing they have both imported the same item(s). 

If we don't get that, could you note it here, please, @valgrimm ?

valgrimm:
will do

stevejalim:
Seems stable again. Closing

PR #1133, opened 2020-01-24
1132 firm up ingestion settings

stevejalim:
This changeset makes the ingestion settings clearer, as part of debugging why stage sent a notification but production did not (and did not ingest the item it should have done)

(Related to #1132)


developerportal/settings/base.py
```diff
@@ -165,7 +165,7 @@
 EMAIL_HOST_PASSWORD = os.environ.get("EMAIL_HOST_PASSWORD")
 EMAIL_HOST_USER = os.environ.get("EMAIL_HOST_USER")
 EMAIL_PORT = int(os.environ.get("EMAIL_PORT", 587))
-EMAIL_USE_TLS = bool(os.environ.get("EMAIL_USE_TLS", True))
+EMAIL_USE_TLS = os.environ.get("EMAIL_USE_TLS", "True") == "True"
```
stevejalim: This is cleaning up a bug where `bool()` doesn't correctly respect an env var being set to `False`, because `bool("False")` is truthy

stevejalim: However if the env var is NOT set, then the default value was being used fine. The change covers the situation where we want to set it to `False` via env vars, which would have failed prior to this change

PR #1150, opened 2020-01-28
1143 + 1144: Task completion survey (core implementation)

stevejalim:
## DO NOT MERGE TO MASTER YET - PLEASE REVIEW, BUT  THIS REQUIRES A SURVEY TO BE SET UP BEFORE PRODUCTION

This changeset adds an initial/MVP version of a prompt that asks the user to complete task-completion survey

To achieve this we:

* show a notification bar
* to 5% of visitors (configurable via settings)
* including a link to a new page (configurable via settings)
* which shows a survey the user can fill in to tell us about their experience

It uses client-side JS to handle selecting the 5% of users and setting a cookie to remember that decision, as well as whether the user has chosen to close the notification. (We looked at using django-waffle to control its appearance trigger, as per MDN, but found we had to make too many compromises to avoid poisoning the CDN.)

Re the CDN: note that at the moment, it's still set to `Vary: Accept-Encoding,Cookie`, which isn't great given that the client-side JS sets a cookie that then computes a new CDN cache key for another cached response that's identical to what we get back without that cookie in a CDN cache key. However, remedying that is ticketed as #1167.

Also note that the NotificationBar component from Mozilla Protocol is currently rendered into the base HTML as a `hidden` node, and then revealed via JS. It can be triggered entirely from JS, but until my [enhancement PR](https://github.com/mozilla/protocol/pull/534) is merged, it's not possible to add a hyperlink to the notification's content via JS. So pre-rendering was the way forward.

(Related issues #1143 and #1144)

## How to test

- check out the site locally or - **ideally** - use the [dev server via its CDN](https://developer-portal-cdn.dev.mdn.mozit.cloud/) 
- if local, in your `local.py` set `TASK_COMPLETION_SURVEY_URL` to something, eg: `"https://example.com/task-completion-survey"`  and `TASK_COMPLETION_SURVEY_PERCENTAGE` to `Decimal("99.9")`
- if on the dev server, note that it's been configured to be shown to 95% of people, and once you get shown it or not shown it, it will stay that way because a cookie is set. Feel free to zap your cookies.
- load the homepage (or any site page) and you will see the notification bar, with a link to the test URL you configured above. This will open a new tab/window for the URL if you click it
- clicking the close `X` on the notification will dismisss it (and it will not be re-shown unless you edit the `devportal_show_task_completion_survey` cookie to be `true`, or you wipe that cookie)

Cookie on load, triggering the message to be shown:
Screenshot 2020-02-05 at 14 03 26

[...]

Issue #1144, opened 2020-01-27
Task completion survey: refine UX by not re-asking 


stevejalim:
**"As a website visitor, I do not want to be shown a feedback questionnaire option for X weeksafter I have already completed one, because that would be annoying."**

(Suggested val for X: 6) 

Acceptance Criteria:
- [ ] Once a user has agreed to (at least) view the survey (we can’t be sure if they complete it or not), they are excluded from the 5% for a period of 6 weeks

Implementation notes:
- This will need cookies to be working via the CDN, so that we can know if a user has been on the site before - cc @limed

Issue #1134, opened 2020-01-24
Consider throttling invalidations


stevejalim:
If someone goes and publishes a lot of new content in rapid succession (eg approving auto-generated posts ingested from another Mozilla blog), then we can face CDN invalidation issues:

https://sentry.prod.mozaws.net/operations/developer-portal-stage/issues/7146814/?query=is:unresolved

> `TooManyInvalidationsInProgress` 
> `developerportal.apps.taskqueue.tasks.invalidate_entire_cdn
> An error occurred (TooManyInvalidationsInProgress) when calling the CreateInvalidation operation: Processing your request will cause you to exceed the maximum number of in-progress wildcard invalidations.

We need to think about this and ways to avoid triggering this situation.

PR #1161, opened 2020-02-03
1144: Task completion survey: mute the prompt if the user dismisses it

stevejalim:
(You should have reviewed PR #1150 before looking at this one, because this builds onthat) 

This changeset adds code that makes use of the django-waffle cookie that determines whether a survey prompt should be shown or not. If its value is True, that means the survey is shown to the user. If it is False, the survey is not shown, but the cookie persists until its expiry date.

Now, when a user clicks the "X" to close/dismiss the notification, this rewrites the value stored in the cookie, to be False, not True. It also has to set a new expiry time, else it'll only exist for the life of the browser session, so we set it to six weeks from now, meaning the user will have a cookie that says "Don't ask me about a survey" for six weeks, after which time it will expire and the user _might_ then (depending on the percentage settings of the survey waffle flag on the server) see a subsequent request to complete a survey - but it will at least have been six weeks.

Note that the decision was taken to ONLY tweak the cookie when the notification is dismissed, and not when the user clicks on the link to view the survey. This seemed to be more sensible in terms of giving users a chance to lose the link/tab and choose to view it again, while also allowing them to expressly dismiss the notificaition.

(Related issue #1144)

## How to test

This will work locally (if you set `WAFFLE_SECURE=False` in your `settings/local.py`) but is also deployed to the [CDN for the dev env](https://developer-portal-cdn.dev.mdn.mozit.cloud/), so everyone can try it.

- zap your cookies for the domain, or use private browsing
- visit the site and confim you see the survey-request notification (it's being shown to 99.9% of visitors, so you should see it)
- in devtools, view your cookies - you'll see `dwf_show_task_completion_survey` is one, with a value of `True`. Note the expiry time.
Screenshot 2020-02-03 at 17 50 52

- dismiss the notification with the `X`
Screenshot 2020-02-03 at 14 27 35

- reload the page
- confirm that the survey-request dialog is no longer shown
- view your cookies again, note that the waffle one now has a value of `False` and that its expiry is six weeks from now.
Screenshot 2020-02-03 at 17 51 00

- also inspect the new/updated cookie and share your views on whether additional settings/parameters need to be set on it, too.

### DIFF
--- a/developerportal/templates/survey.html
+++ b/developerportal/templates/survey.html
@@ -16,7 +16,7 @@
   <button class="mzp-c-notification-bar-button mzp-js-notification-trigger" type="button"></button>
   <p>
       Your feedback is important. Would you
-        <a target="_blank" rel="noreferrer noopener" href="{{survey_url}}" class="mzp-c-notification-bar-cta">
+        <a target="_blank" rel="noreferrer noopener" href="{{survey_url}}" class="mzp-c-notification-bar-cta js-survey-link">
           complete a short survey
         </a>
       after visiting?
--- a/src/js/task-completion-prompt.js
+++ b/src/js/task-completion-prompt.js
@@ -1,27 +1,92 @@
 /* eslint-disable prefer-template */
-const { getCookie } = require('./utils');
+const { getCookieValue, setCookieValue } = require('./utils');
+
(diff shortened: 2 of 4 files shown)


stevejalim:
Closing because heavily refactoring #1134 / #1150

Issue #1203, opened 2020-03-09
Support ContentPage children of Topic Pages (and all other pages)


stevejalim:
To support the appropriat structure of the Topic/Product landing pages and their descendents, we need to allow Topic Pages to have ContentPage children.

PR #1205, opened 2020-03-09
1203: Allow Topic pages to have a ContentPage as their child page.

stevejalim:
This will make the entire landing/products and technology heirarchy easier to set up the way we need.

A topic page now allows the addition of a Content Page as a child:

Screenshot 2020-03-09 at 17 36 57

And the child page's URL path sits beneath that of the parent Topic page:

Screenshot 2020-03-09 at 17 38 35

## How to test
Please try it out on the dev server

https://developer-portal.dev.mdn.mozit.cloud/topics/test-topic/test-subpage-topic-page/

(Related issue #1203)

### DIFF
--- a/developerportal/apps/content/models.py
+++ b/developerportal/apps/content/models.py
@@ -23,7 +23,7 @@ class ContentPageTag(TaggedItemBase):
 
 
 class ContentPage(BasePage):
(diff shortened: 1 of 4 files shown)


stevejalim:
Note that this doesn't fix the page hanging issue, but DOES let us have a more appropriate 'family tree' of pages

Issue #1207, opened 2020-03-10
Support a button-styled hyperlink as a rich-text panel item


stevejalim:
As part of an effort to tackle mysteriously hanging pages, we should try to add a button-styled hyperlink as a rich-text panel item. This would mean that we could reduce the number of blocks in a StreamField, which should result in less intensive content structures.

Acceptance criteria:

- [ ] all rich-text editor panels in the CMS will have a new option beside the hyperlink (possibly called "Button") which will work like a hyperlink option when adding one
- [ ] the Wagtail editor **should** render these items as a button-styled link, if possible
- [ ] the actual previewed or published pagess **MUST** render these items as button-styled links

stevejalim:
cc @valgrimm for awareness

PR #1215, opened 2020-03-10
1207: Add support for styling specified links as buttons in Wagtail's rich-text editor

stevejalim:
(Note: a big hat-tip goes to @kalobtaulien for his assistance with this, making it a much smaller change than writing an entirely new Draftail Entity. \o/ )

This changeset supports a way of marking certain links in a rich-text area as buttons, so that they are rendered as such in the published page.

(This should allow us to make pages that feature lots of buttons without the page becoming too heavy/complex, which appears to be a factor in editing hanging after a certain tipping point.)

It acheives this by adding a new button to the Draftail editor which adds a new block (a div, lightly styled with a dashed grey border) inside which one can add hyperlinks using rich-text editor's standard link button. Any links inside this block will be rendered as buttons in the published (and previewed) page, and are lightly styled to look similar within the rich-text editor itself.

Editor:
Screenshot 2020-03-10 at 21 26 45

Rendered:
Screenshot 2020-03-10 at 21 16 50

@valgrimm This changeset doesn't 100% sort out the spacing-after-buttons issue, but we can address that separately - eg with a button to add a "spacer" block if we really want it, or increase the default spacing after a rich-text area (or just after one of these new button blocks).

(Related issue #1207)

## How to test

- once this code is staged on dev, add/edit a Post or any other page with a full rich-text editor. Click the icon to add a `button-block` block and add some links inside it.
- Here's [a page](https://developer-portal.dev.mdn.mozit.cloud/topics/test-topic/example-page-links-blocks-buttons/) i made to test on dev, too

### DIFF
--- a/developerportal/apps/common/constants.py
+++ b/developerportal/apps/common/constants.py
@@ -49,6 +49,7 @@ RICH_TEXT_FEATURES = (
     "image",
     "italic",
     "link",
+    "button-block",  # custom - see common/wagtail_hooks.py
     "ol",
     "ul",
 )
--- a/developerportal/apps/common/wagtail_hooks.py
+++ b/developerportal/apps/common/wagtail_hooks.py
@@ -1,6 +1,9 @@
 # pylint: disable=no-member
-from django.utils.html import escape
+from django.contrib.staticfiles.templatetags.staticfiles import static
+from django.utils.html import escape, format_html
 
+import wagtail.admin.rich_text.editors.draftail.features as draftail_features
+from wagtail.admin.rich_text.converters.html_to_contentstate import BlockElementHandler
 from wagtail.core import hooks
 from wagtail.core.rich_text import LinkHandler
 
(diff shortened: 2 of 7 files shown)


Gregoor:
> (This should allow us to make pages that feature lots of buttons without the page becoming too heavy/complex, which appears to be a factor in editing hanging after a certain tipping point.)

This is throwing me off a bit. How does visually turning links into buttons make the page lighter?

Otherwise this looks good!

stevejalim:
@Gregoor 
> This is throwing me off a bit. How does visually turning links into buttons make the page lighter?

The links are only styled as buttons (so don't add extra markup) and this approach means we don't have to:

````
1 stop adding content in a rich-text editor
2 add a button element
3 add a new rich-text editor to add more content
4 GOTO 1
```
because adding so many extra elements to the StreamField creates a large payload that's causing us hurt elsewhere.

Does that help explain it a bit?

Gregoor:
> 
> 
> @Gregoor
> 
> > This is throwing me off a bit. How does visually turning links into buttons make the page lighter?
> 
> The links are only styled as buttons (so don't add extra markup) and this approach means we don't have to:
> 
> ```
> 1 add content by adding a new rich-text editor to the streamfield
> 2 add a button element to the streamfield
> 3 add a new rich-text editor to the streamfield to add more content
> 4 GOTO 2
> ```
> 
> because adding so many extra elements to the StreamField creates a large payload that's causing us hurt elsewhere.
> 
> Does that help explain it a bit?

Thanks Steve, it did. I didn't fully understand those streamfield implications (and probably still have some catching up to do there).

Issue #1213, opened 2020-03-10
Move gunicorn config to a file instead


limed:
The options called by gunicorn via the cli is starting to get messy, so lets start using a config file instead

Issue #1211, opened 2020-03-10
Tweak  gunicorn config to use /dev/shm for worker_tmp_dir


limed:
See: https://docs.gunicorn.org/en/stable/faq.html#how-do-i-avoid-gunicorn-excessively-blocking-in-os-fchmod

Specifically it says

`
 AWS an EBS root instance volume may sometimes hang for half a minute and during this time Gunicorn workers may completely block in os.fchmod. os.fchmod may introduce extra delays if the disk gets full. Also Gunicorn may refuse to start if it can’t create the files when the disk is full.
`

So we should configure `--worker-tmp-dir` to point to `/dev/shm`

stevejalim:
@limed Still a thing worth doing?

limed:
Already doing this since #1245 is merged

PR #1245, opened 2020-03-12
Move gunicorn config to its own file

limed:
This moves the gunicorn config to a config file instead of calling
command line options. This will only affect local dev for now, and there
should be another PR to actually have kubernetes use the config file as
well

(Related issue #1213 & #1211)

## Key changes:

- Moves gunicorn config to a config file

### DIFF
--- a/configs/gunicorn.py
+++ b/configs/gunicorn.py
@@ -1,0 +1,37 @@
+from os import getenv
+from multiprocessing import cpu_count
+
+bind = f"0.0.0.0:{getenv('APP_PORT', '8000')}"
+
+# Log section
+if getenv('DEBUG') == "true":
+    loglevel = 'debug'
+else:
+    loglevel = 'info'
+
+accesslog = '-'
+errorlog = '-'
+access_log_format = '%(h)s %(l)s %(u)s %(t)s \"%(r)s\" %(s)s %(b)s \"%(f)s\" \"%(a)s\" %({X_Forwarded_For}i)s"'
+
+# How long to keep connection alive, this should be timed
+# a little less than the ELB idle timeout
+keepalive = getenv('APP_GUNICORN_TIMEOUT', '118')
+
+# NOTE: use keepalive instead of timeout
+# timout = getenv('APP_GUNICORN_TIMEOUT', '118')
+
+if getenv('DJANGO_ENV') == 'dev':
+    workers = '3'
+else:
+    # Automatically figure number of workers if environment is not set
+    cpu = (cpu_count() * 2 + 1)
+    workers = getenv('APP_GUNICORN_WORKERS', cpu)
+
+worker_connections = getenv('APP_GUNICORN_WORKER_CONNECTIONS', '1000')
+worker_class = getenv('GUNICORN_WORKER_CLASS', 'gunicorn.workers.ggevent.GeventWorker')
+worker_tmp_dir = '/dev/shm'
+
+
+# Called just after a worker has been forked.
+def post_fork(server, worker):
+    server.log.info("Worker spawned (pid: %s)", worker.pid)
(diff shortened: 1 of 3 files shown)


configs/gunicorn.py
```diff
@@ -0,0 +1,37 @@
+from os import getenv
+from multiprocessing import cpu_count
+
+bind = f"0.0.0.0:{getenv('APP_PORT', '8000')}"
+
+# Log section
+if getenv('DEBUG') == "true":
```
stevejalim: Worth a tweak to handle `True` and `true` in the config

```suggestion
if getenv('DEBUG').lower() == "true":
```

configs/gunicorn.py
```diff
@@ -0,0 +1,37 @@
...
+accesslog = '-'
+errorlog = '-'
+access_log_format = '%(h)s %(l)s %(u)s %(t)s \"%(r)s\" %(s)s %(b)s \"%(f)s\" \"%(a)s\" %({X_Forwarded_For}i)s"'
+
+# How long to keep connection alive, this should be timed
+# a little less than the ELB idle timeout
+keepalive = getenv('APP_GUNICORN_TIMEOUT', '118')
+
+# NOTE: use keepalive instead of timeout
+# timout = getenv('APP_GUNICORN_TIMEOUT', '118')
```
stevejalim:
```suggestion
# timeout = getenv('APP_GUNICORN_TIMEOUT', '118')
```

configs/gunicorn.py
```diff
@@ -0,0 +1,37 @@
...
+else:
+    loglevel = 'info'
+
+accesslog = '-'
+errorlog = '-'
+access_log_format = '%(h)s %(l)s %(u)s %(t)s \"%(r)s\" %(s)s %(b)s \"%(f)s\" \"%(a)s\" %({X_Forwarded_For}i)s"'
+
+# How long to keep connection alive, this should be timed
+# a little less than the ELB idle timeout
+keepalive = getenv('APP_GUNICORN_TIMEOUT', '118')
```
stevejalim: Does this need to be an int or OK as a string?

configs/gunicorn.py
```diff
@@ -0,0 +1,37 @@
...
+
+# NOTE: use keepalive instead of timeout
+# timout = getenv('APP_GUNICORN_TIMEOUT', '118')
+
+if getenv('DJANGO_ENV') == 'dev':
+    workers = '3'
+else:
+    # Automatically figure number of workers if environment is not set
+    cpu = (cpu_count() * 2 + 1)
+    workers = getenv('APP_GUNICORN_WORKERS', cpu)
```
stevejalim: Similarly, should this be a string (which it would be if it's in the env as `APP_GUNICORN_WORKERS`) or an int?

configs/gunicorn.py
```diff
@@ -0,0 +1,37 @@
...
+
+worker_connections = getenv('APP_GUNICORN_WORKER_CONNECTIONS', '1000')
+worker_class = getenv('GUNICORN_WORKER_CLASS', 'gunicorn.workers.ggevent.GeventWorker')
```
stevejalim: 🙌

configs/gunicorn.py
limed: It didn't seem to care if it was an int or a string when i tested it out

configs/gunicorn.py
limed: Same as above and it didn't seem to really care about it being an int or string

configs/gunicorn.py
limed: Fixed in 1ad3b8542bdb2162b82a72c3d5dbd68d51bd7b7f

Issue #1224, opened 2020-03-11
Update newsletter signup component to new design


stevejalim:
- [ ] UI updates
- [ ] JS updates to post to it

valgrimm:
When it is testing time, this behavior is worth investigating

![image.png](https://images.zenhubusercontent.com/5db08cf412d6350001ab877a/18acf715-b11d-4412-9e48-ff0a52bb05a1)

valgrimm:
Consider this stakeholder approved as no one complained about anything other than header page last Monday. All there is blocking completion is that bug

PR #1246, opened 2020-03-12
1224: Design refresh - Newsletter signup panel

stevejalim:
This changeset refreshes the newsletter signup panel and plumbs in direct signup-in-the-page behaviour, which we didn't have before.

* The UI is currently using a placeholder image (in the desktop viewport; the image is not displayed on mobile), but otherwise follows the designs from @malqinneh 

Screenshot 2020-03-12 at 22 21 13

Screenshot 2020-03-12 at 22 12 18

* The JS-powered newsletter signup is heavily inspired by what @Gregoor did in Kuma for the same newsletter signup, but because we're not using React or the like, the post-submission behaviour is simpler: either replacing the HTML form with a message, or triggering an alert()

Screenshot 2020-03-12 at 21 59 32

Screenshot 2020-03-12 at 21 59 12

(Related issue #1224)

## How to test

- will be staged on dev server soon


src/js/tests/newsletter-subscription.test.js
```diff
@@ -0,0 +1,95 @@
...
+  });
+  expect(fetch.mock.calls[0][1].method).toEqual('POST');
+  expect(fetch.mock.calls[0][1].body).toEqual(
+    'newsletters=app-dev&fmt=H&email=test%40example.com',
+  );
+  expect(fetch.mock.calls[0][0]).toEqual(
+    'https://www.mozilla.org/en-US/newsletter/',
+  );
+
+  expect(fetch.mock.calls.length).toEqual(2);
```
stevejalim: @Gregoor This value should really be `1` in my opinion, unless `fetch` does some double-posting as normal...?

If you run the project locally and look at the network panel when you sign up, you'll see the browser makes two requests, too, so it's not only in Jest that it works (and note that in Jest there is only one DOM element that matches the selector used for binding, etc). So, suggestions about what I've got wrong here would be welcome.

Once I'm clearer about why it's 2 / how to make it work as expected with just 1 POST, I can add some unhappy paths to the tests too.

src/js/tests/newsletter-subscription.test.js
```diff
@@ -0,0 +1,95 @@
...
+ * global to call 'fetchMock' instead of the real implementation
+ */
+
+/* global fetchMock */
+
+require('jest-fetch-mock').enableMocks();
+
+Object.defineProperty(global.window, 'alert', function fakeAlert(val) {
+  // eslint-disable-next-line no-console
+  console.log(val);
```
stevejalim: Note: I'll be replacing this with a `spy`

src/js/tests/newsletter-subscription.test.js
stevejalim: Screenshot 2020-03-13 at 11 59 00

src/js/newsletter-subscription.js
```diff
@@ -0,0 +1,87 @@
...
+            return;
+          }
+
+          if (errors && errors.length) {
+            complain(`There was a problem subscribing you: ${errors}`);
+          }
+        })
+        .catch(() => {
+          complain(`There was a problem subscribing you. Please try again.`);
+          // TODO? ping monitoring with ([e.toString()]) ?
```
Gregoor: probably indeed worth adding sentry error logging here (and in general for frontend errors)

src/js/newsletter-subscription.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return null; // just to keep linter happy
+  }
+
+  constructor(form) {
+    this.form = form;
+    this.form.onsubmit = event => {
+      event.preventDefault();
+
+      submitNewsletterSubscription(this.form)
```
Gregoor: can't we use async-await? Tends to make things more readable.

src/js/tests/newsletter-subscription.test.js
```diff
@@ -0,0 +1,91 @@
...
+  expect(fetch.mock.calls[0][1].method).toEqual('POST');
+  expect(fetch.mock.calls[0][1].body).toEqual(
+    'newsletters=app-dev&fmt=H&email=test%40example.com',
+  );
+  expect(fetch.mock.calls[0][0]).toEqual(
+    'https://www.mozilla.org/en-US/newsletter/',
+  );
+
+  // show the content in js-newsletter-fields has been updated after a successful POST
+  // THIS IS CURRENTLY FAILING EVEN THOUGH IT WORKS IN THE REAL BROWSER
```
stevejalim: @gregor This currently failing because the js-dom(?) component isn't being updated as far as the test is concerned, even though adding console.log(document.getElementsByClassName('js-newsletter-fields')[0].innerHTML); adds output to the test runner that shows it HAS been udpdated. Odd.

stevejalim: All brainwaves welcome!

malqinneh:
Looks great!

PR #1255, opened 2020-03-13
1224 newsletter signup update

stevejalim:
*REOPENED after premature merge*

Already approved by @malqinneh  and @valgrimm  but re-opened for @Gregoor's laser eyes

---

This changeset refreshes the newsletter signup panel and plumbs in direct signup-in-the-page behaviour, which we didn't have before.

* The UI is currently using a placeholder image (in the desktop viewport; the image is not displayed on mobile), but otherwise follows the designs from @malqinneh 

Screenshot 2020-03-12 at 22 21 13

Screenshot 2020-03-12 at 22 12 18

* The JS-powered newsletter signup is heavily inspired by what @Gregoor did in Kuma for the same newsletter signup, but because we're not using React or the like, the post-submission behaviour is simpler: either replacing the HTML form with a message, or triggering an alert()

Screenshot 2020-03-12 at 21 59 32

Screenshot 2020-03-12 at 21 59 12

(Related issue #1224)

## How to test

- will be staged on dev server soon


src/js/tests/newsletter-subscription.test.js
```diff
@@ -0,0 +1,91 @@
...
+  expect(fetch.mock.calls[0][1].method).toEqual('POST');
+  expect(fetch.mock.calls[0][1].body).toEqual(
+    'newsletters=app-dev&fmt=H&email=test%40example.com',
+  );
+  expect(fetch.mock.calls[0][0]).toEqual(
+    'https://www.mozilla.org/en-US/newsletter/',
+  );
+
+  // show the content in js-newsletter-fields has been updated after a successful POST
+  // THIS IS CURRENTLY FAILING EVEN THOUGH IT WORKS IN THE REAL BROWSER
```
stevejalim: @Gregoor This, when uncommented, is currently failing because the js-dom(?) `document.body` isn't being updated as far as the test is concerned, even though adding `console.log(document.getElementsByClassName('js-newsletter-fields')[0].innerHTML);` to the app code adds output to the test runner that shows it HAS been udpdated. Odd.

Gregoor: Wasn't sure if I could commit here, so here's a patch for you. Problem is that the DOM change happen in the next tick of the event loop. There might be a more elegant way to make sure the assert happens after, but I can't seem to find one right now and this does work (as long as whatever is async doesn't take more than a tick):

```diff
Index: src/js/tests/newsletter-subscription.test.js
<+>UTF-8
===================================================================
--- src/js/tests/newsletter-subscription.test.js	(revision 9611cdb0d8ca61117f0efafde31bc3614071e690)
+++ src/js/tests/newsletter-subscription.test.js	(date 1584383512698)
@@ -53,7 +53,7 @@
   fetch.resetMocks();
 });
 
-test('Happy path', () => {
+test('Happy path', async () => {
   fetchMock.mockResponseOnce(JSON.stringify({ success: true }));
   document.body.innerHTML = exampleFormHTML;
   NewsletterSubscription.init(); // this is how it's invoked in index.js
@@ -79,13 +79,13 @@
     'https://www.mozilla.org/en-US/newsletter/',
   );
 
+  await new Promise(resolve => setTimeout(resolve));
+
   // show the content in js-newsletter-fields has been updated after a successful POST
-  // THIS IS CURRENTLY FAILING EVEN THOUGH IT WORKS IN THE REAL BROWSER
-  //
-  // const newsletterFieldsDivContent = form.getElementsByClassName(
-  //   'js-newsletter-fields',
-  // )[0].innerHTML;
-  // expect(newsletterFieldsDivContent).toBe(
-  //   '<b>Thank you. Please check your email to confirm your subscription.</b>',
-  // );
+  const newsletterFieldsDivContent = form.getElementsByClassName(
+    'js-newsletter-fields',
+  )[0].innerHTML;
+  expect(newsletterFieldsDivContent).toBe(
+    '<b>Thank you. Please check your email to confirm your subscription.</b>',
+  );
 });
```

stevejalim:
Final version with the real pic in it:

Screenshot 2020-03-17 at 21 49 26

Issue #1220, opened 2020-03-11
Update Homepage


stevejalim:
Acceptance criteria

- [ ] width of the site matches Mozilla.org  
- [ ] grid-gutters, etc match Mozilla.org  
- [ ] Homepage reflects design from Mustafa
- [ ] Page passes HTML validation

valgrimm:
Consider this stakeholder approved as no one complained about anything other than header page last Monday

PR #1274, opened 2020-03-17
1220 Design refresh: homepage

stevejalim:
This changeset update the homepage to use the new designs (which are basically following the mozilla.org / core Protocol UI cues).

It's a large changeset, unfortunately, because it included amending some core concepts such as the Cards. However, the upside is that this will make later pages faster to update.

(Note that the Homepage passes HTML validation checks)

(Related issue #1220)
## How to test

- I've staged it on the dev server (DM me for a URL if you need it)

# Cross-browser checked:

- [X] IE 11 Win 10 -
    - OK but newsletter layout fails in IE11
- [X] Edge Win 10
- [X] FF Win
- [X] FF Mac
- [X] Chrome Win
- [X] Chrome Mac
- [X] Safari (5.1) Win
  - visually not great at all, but not a core browser
- [X] Safari Mac
- [X] Android Chrome
- [X] Mobile Safari

## Screenshot 

(Ignore the newsletter signup panel - that's on a different branch)

Screenshot_2020-03-17 Developer tools, resources, and more


stevejalim:
Reopening because not merged yet

stevejalim:
@malqinneh can you feed back your tweaks and we'll put them in a follow-up ticket, please?

developerportal/templates/molecules/card-featured.html
```diff
@@ -1,27 +1,54 @@
...
-    <div class="card-featured-image" style="background-image: url('{% firstof card_image.url image.url fallback_image_url %}')"></div>
-    <div class="card-featured-caption">
-      <h5 class="no-underline">
-        {% firstof resource.card_title resource.title %}{% if external_page or resource.is_external %}<span class="no-wrap">&#65279;<span class="icon icon-external">{% include "atoms/icons/external.svg" %}</span></span>{% endif %}
-      </h5>
+  <div class="mzp-c-card-media-wrapper">
+      <div class="lazy-image-container">
+        <img class="mzp-c-card-image" src="{% firstof img.url fallback_url %}" alt="">
+      </div>
+  </div>
```
schalkneethling: General question: Why do we need two wrapping `div` elements for the one `img`. I guess the CSS framework might be enforcing that.

schalkneethling: General question: Why do we need two wrapping `div` elements for the one `img`. I guess the CSS framework might be enforcing that.

developerportal/templates/molecules/item-person.html
```diff
@@ -1,39 +1,39 @@
...
-            <div class="person-title highlight1">
-              {{ person.title }}
-            </div>
-          {% endif %}
+      <div class="mzp-c-card-content">
+        <div class="mzp-c-card-tag">{{person.get_role_display}}</div>
+        <h2 class="mzp-c-card-title">
+          <span>
+            {{ person.title }}
+          </span>
```
schalkneethling: Curious about the need to wrap this in a `span` tag. Looking the page, it does not look there is any CSS added based on the nested `span`

developerportal/templates/molecules/item-topic-link.html
```diff
@@ -16,4 +19,18 @@ <h4 class="no-underline">{{ topic.title }}</h4>
       {{ topic.description | richtext }}
     {% endif %}
   </div>
-</a>
+</a> {% endcomment %}
```
schalkneethling: Is everything between `{% comment %}` and `{% endcomment %}` excluded from what is sent to the browser? Or are these not that type of comment :)

developerportal/templates/molecules/item-topic-link.html
```diff
@@ -16,4 +19,18 @@ <h4 class="no-underline">{{ topic.title }}</h4>
...
+
+<section class="mzp-c-card mzp-c-card-extra-small">
+  <a class="mzp-c-card-block-link" href="{{ topic.url }}">
+    <div class="mzp-c-card-content">
+      <img src="{% get_menu_item_icon topic %}" class="topic-item-icon" alt="" width="24" height="24">
+      <h2 class="mzp-c-card-title"><span>{{ topic.title }}</span></h2>
+      {% if topic.description %}
+        <div class="mzp-c-card-desc">
+        {{ topic.description | richtext }}
+        </div>
```
schalkneethling: Would it not be more semantic for this to be marked up as a `p` tag?

stevejalim: Thanks for the suggestion.

The problem here is that the `richtext` filter outputs a `div` and/or `p` depending on what's in it, so we can't risk a wrapping `p`, alas

developerportal/templates/molecules/item-topic-link.html
stevejalim: Yep, stuff inside `{% comment %}` is excluded, but this has since been amended/reverted/updated

developerportal/templates/molecules/item-person.html
stevejalim: This file is no longer present in the codebase - was refactored away

developerportal/templates/molecules/card-featured.html
stevejalim: Good question. Answer: am going with the reference implementations in Protocol. 

(Personally, I find it easier to scan, too, fwiw)

Issue #1266, opened 2020-03-17
Design refresh: Events page 


stevejalim:

stevejalim:
@malqinneh @valgrimm I was just thinking about this page and how we will be switching to adding `Past events` option to the `Dates` section of the filters.

If we wanted to give users a quick way to access past events, we could also use an ad-hoc button that currently appears below the Featured Event but above the listings as a short-cut to enable the 'past events' filter, if you see what I mean

What do you think?

valgrimm:
Note: We current plan to keep functionality from current events page to add buttons above main content as requested by Sandra

![image.png](https://images.zenhubusercontent.com/5db08cf412d6350001ab877a/d623f526-f4df-4117-bbac-fcda8abd7984)

PR #1285, opened 2020-03-19
1266: Events page design refresh (design-only pass)

stevejalim:
This changeset updates the look of the Events listing page (`/events/`) as per the new designs.

It does not include mixing the past events into the results - that's carded separately and will come soon.

It adds a new single-featured-item card layout with a 3:1 image, which is not in Mozilla Protocol, but is needed in Devportal.

(Related issue #1266)

## Screenshots

Screenshot_2020-03-19 Events(2)

----

Screenshot_2020-03-19 Events(3)

## How to test

- Up on dev server for experimentations and test driving.


stevejalim:
@valgrimm @malqinneh I know there are no pagination controls on this yet, but I will add them when i circle back to sort the past events filter

stevejalim:
Just noticed I missed the black header strip too - will add that shortly

stevejalim:
OK, black header strip added and screenshots updated. Pushing to dev server now

stevejalim:
@malqinneh One thing I've noticed is that Protocol limits the max width of the black header strip -- take a look at the dev site on a large screen/window and you'll see what i mean. Is that OK with you?

developerportal/templates/molecules/card-featured-primary.html
```diff
@@ -20,7 +20,7 @@
 
 
 <div class="mzp-l-card-full">
-  <section class="mzp-c-card mzp-c-card-x-large mzp-has-aspect-16-9">
+  <section class="mzp-c-card devportal-card-x-large devportal-has-aspect-3-1">
```
schalkneethling: In a lot of other places(not just this PR), I see `section` elements changed to `div`. That is not the case here. Is there a particular reason to keep this one a `section`?

developerportal/templates/molecules/cards/card-event.html
```diff
@@ -3,69 +3,32 @@
...
-    rel="noopener noreferrer"
-  {% endif %}
-  data-type="{{ resource.resource_type }}"
->
-  <div class="card card-event{% if featured %} card-event-featured{% endif %}">
-    <div class="card-event-ribbon-mzp-l-content">
-      <div class="card-event-ribbon background-accent">
-        <span class="highlight2">{{ resource.event_dates|safe }}</span>
-      </div>
+<section class="mzp-c-card mzp-c-card-medium mzp-has-aspect-16-9">
```
schalkneethling: Also here, a specific reason to keep this a section? Generally, the `section` element is of very little value semantically unless it is being used to define an actual `region`. See this great writeup https://www.smashingmagazine.com/2020/01/html5-article-section/#so-what-about-lt-section-gt

developerportal/templates/organisms/filter-list.html
```diff
@@ -1,54 +1,88 @@
...
-              {% empty %}
-                <div class="filter-list-no-results">
-                  <h2>{{ no_resources_message|default:"No results found" }}</h2>
-                </div>
-              {% endfor %}
-            </div>
+      {% empty %}
+        <div class="filter-list-no-results">
+          <h2>{{ no_resources_message|default:"No results found" }}</h2>
+        </div>
```
schalkneethling: Could we just use the `h2` or do we need the wrapping `div` element?

src/css/global/layout.scss
```diff
@@ -0,0 +1,11 @@
+// Override Protocol's 25% / 75% layout to be 33% / 67%
+
+@media #{$mq-md} {
+  .mzp-l-sidebar {
+    width: 33.333% !important;
```
schalkneethling: Is this to style this container? https://github.com/mdn/developer-portal/pull/1285/files#diff-b7ede9403e1fba8bb7a9047a07b700e1R5

If so, could we do:

```
aside.mzp-l-sidebar {
  width: 33.333%;
}
```

src/css/global/layout.scss
```diff
@@ -0,0 +1,11 @@
...
+  }
+
+  .mzp-l-main {
```
schalkneethling: Looks like that is meant for this container:
https://github.com/mdn/developer-portal/pull/1285/files#diff-b7ede9403e1fba8bb7a9047a07b700e1R15

Could we add a custom class and in so doing avoid the use of `!important`, for example:


```
<div class="mzp-l-main custom-width">
...

.mzp-l-main.custom-width {
   width: 66.665%;
}
```

src/css/molecules/card-featured.scss
```diff
@@ -1,182 +1,25 @@
...
+.mzp-c-card.devportal-has-aspect-3-1 .mzp-c-card-media-wrapper iframe,
+.mzp-c-card.devportal-has-aspect-3-1 .mzp-c-card-media-wrapper object,
+.mzp-c-card.devportal-has-aspect-3-1 .mzp-c-card-media-wrapper embed {
+  bottom: 0;
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: -50%; // to mock-vertically center the image
+  /*! height:  50%; */
+}
```
schalkneethling: As this is SCSS could we rewrite this as

```suggestion
.mzp-c-card.devportal-has-aspect-3-1 {
  .mzp-c-card-media-wrapper {
    padding-bottom: 33.33%;
    position: relative;
    
    img,
    video,
    iframe,
    object,
    embed {
      position: absolute;
      top: -50%; /* to mock-vertically center the image */
      right: 0;
      bottom: 0;
      left: 0;
    }
  }
}
```

src/css/molecules/card.scss
```diff
@@ -381,6 +381,9 @@
   }
 }
 
+/* -------------------------------------------------------------------------- */
+// NEW CUSTOM CSS FOR CARDS
+
```
schalkneethling: Left overs?

src/css/organisms/filter-list.scss
```diff
@@ -1,25 +1,25 @@
...
+  // @media #{$mq-md} {
+  //   column-gap: $grid-gap;
+  //   grid-template-columns:
+  //     [sidebar-start] repeat(4, 1fr) [sidebar-end items-start] repeat(8, 1fr)
+  //     [items-end];
+
+  //   html.no-js & {
+  //     grid-template-columns: [items-start] 1fr [items-end];
+  //   }
+  // }
```
schalkneethling: Could this be safely deleted or is there a specific reason to keep it around?

src/css/organisms/filter-list.scss
```diff
@@ -1,25 +1,25 @@
...
 }
 
 .filter-list-sidebar {
-  -ms-grid-column: 1;
-  grid-column: sidebar;
+  // -ms-grid-column: 1;
+  // grid-column: sidebar;
```
schalkneethling: Could this be safely deleted or is there a specific reason to keep it around?

src/css/organisms/filter-list.scss
```diff
@@ -68,8 +69,8 @@
 }
 
 .filter-list-items {
-  -ms-grid-column: 2;
-  grid-column: items;
+  // -ms-grid-column: 2;
+  // grid-column: items;
```
schalkneethling: Could this be safely deleted or is there a specific reason to keep it around?

schalkneethling:
@stevejalim What is the browser support matrix for the portal? aka, in which browsers and versions should I be testing? Thank you

developerportal/templates/molecules/card-featured-primary.html
stevejalim: True - the original markup had lots of `section` and I basically am not being consistent enough with replacement. Willdo!

developerportal/templates/organisms/filter-list.html
stevejalim: Tidying this up is on my list for the next pass. 👍

src/css/global/layout.scss
stevejalim: We'll be using that override in other templates too, but it will always be an aside, so I could make it more specific, sure.

(Or are you saying it doesn't need to be in a media query and we just override it everywhere? I didn't do that because I want it to flow as above-and-below in small viewports)

src/css/global/layout.scss
stevejalim: Sure we can do that. I'll add it in on a future pass

src/css/molecules/card-featured.scss
stevejalim: Oh that's much nicer. What is there is based on the generated CSS from Protocol, not the source!

src/css/molecules/card.scss
stevejalim: Yep - i'll be dropping it when am sure we don't need any bits of it. There'll be a dedicated cleanup ticket

stevejalim:
@schalkneethling 

> What is the browser support matrix for the portal? aka, in which browsers and versions should I be testing? Thank you

Am aiming for evergreen FF, Chrome, Safari on Win 10 and Mac Catalina, plus evergreen Edge and IE11 on Win 10 plus latest iOS Safari and Android Chrome

stevejalim:
Amends made @schalkneethling and I'll update the dev server shortly!

schalkneethling:
When using the reponsive devtools in both Firefox and Chrome I see the following when selecting iPad(portrait)

Screen Shot 2020-03-23 at 06 46 51-fullpage

Notice the newsletter section just above the footer. It then causes horizontal scrolling.

Looking at the CSS, it seems to be caused by this piece:

```
@media (min-width:768px) {
 .newsletter-signup {
  display:grid;
  grid-template-columns: 15% repeat(2,1fr) 15%;
  grid-gap:24px
 }
}
```

This suggests 4 columns but the container only contains two child elements. Changing the above to the following seems to resolve the problem:

```
@media (min-width:768px) {
 .newsletter-signup {
  display:grid;
  grid-template-columns: repeat(2,1fr);
  grid-gap:24px
 }
}
```

NOTE: Strangely enough I do not see the same thing happen when testing using either an iPad simulator or an iPad via Browserstack.

schalkneethling:
I can also confirm that I can see the above if I manually resize the browser window so, it seems like it is a responsive problem and on a device such as the iPad it will be all good. Worth making the change anyhow?

schalkneethling:
I also notice that the newsletter section is "broken" in IE11 because it does not support `grid`:

Screenshot 2020-03-23 at 09 02 27

One option is to use `flex` instead and then progressively enhance with `@supports` and grid if there is value to be gained.

If the framework is kinda forcing us to have `display: grid` based on the classes added to the elements, then I guess we will have to override those classes directly and add this behavior. Not sure if this needs to be addressed in the PR or whether we should spin up a separate issue for that.

Let me know @stevejalim

stevejalim:
@schalkneethling Thanks for all the feedback! I'll sort out the grid issue shortly.

Re IE11: so far we've gone with toleratig visual imperfections, including reflow issues, as long as the content is readable and usable. The newsletter flow problem could be seen as "just" that, but if you don't mind spinning up another card re using flexbox instead, then I can look at that when I get time, because it would be nice to remedy. Cheers!

stevejalim:
@schalkneethling Updated to use flexbox instead and it's a lot happier

**FF happy at zoomed-out/low magnification**
Screenshot 2020-03-23 at 15 59 38

**IE11 now happy**
Screenshot 2020-03-23 at 15 59 07

**Mobile viewport still happy**
Screenshot 2020-03-23 at 16 04 33


( This also resolves #1290 )

Issue #1288, opened 2020-03-19
Design refresh: Support past events on Events page


stevejalim:
AC 

- [ ] "Past events" is an option at the bottom of the filter list
- [ ] Selecting it includes ALL past events along with any chosen by other selected filters

(Also we should remove redundant code that looks up past events by default to make page loading optimal - small task)

stevejalim:
I couldn't find the right card that captured this requirement so made a new card.

//cc @valgrimm

PR #1293, opened 2020-03-21
1288: Events page -- include past events

stevejalim:
This changeset adds support for past events as an option in the Events-page sidebar filtering. It also adds pagination to the events list, which was absent from the design-focused #1266.

Screenshot_2020-03-23 Events

When selected, the results will be scoped to all appropriate past events (further filtered by any additional options checked, such as topic or region). If other dates are selected, matching Events will be shown, too.

This changeset also reverses the ordering of the events show, so that we now have the newest first. This is to stop the old events dominating the results when 'include past events' is selected. The order of the dates in the filter panel also reflects this.

Also, the Country list has been expanded to include all places for past events, too, else it would not be possible to filter by all relevant countries when 'include past events' is ticked. 

(Related issue #1288 )

## How to test

- The code is staged on dev. Please try out the various filter groups. eg:
  -  if ONLY 'past events' in Dates is selected, it scopes to all past events in the results, including applying any relevant filters to it (so you can find future and past "CSS" events, for instance)
  - if you filter by region or P&T and ONLY have past events selected, you ONLY get to see past events


stevejalim:
@malqinneh Yeah, I get you and it definitely makes more sense to include them by default and have the filters 'fine tune' what's included. Will be a quick change, so getting started on it now

stevejalim:
@malqinneh 

> Also, can we anchor 'Past events' at the top of the [Date] list?

I just tried this out then realised that if the result list is running newest -> oldest, it makes sense for the sidebar filter options to match, so have kept it at the bottom but changed the order of the other dates. Once it's up on dev (Monday AM your time), let me know what you think. I'll update screenshots here, too

stevejalim:
@valgrimm @malqinneh updated on dev - let me know what you think

@valgrimm I'm noticing some card images on dev aren't great in terms of fit/flow/sizing now we're using the new card cSS - something to discuss in Slack?

valgrimm:
> 
> 
> Apologies if conversations already happened about this but was thinking past events would be included in the events results when user(s) land on the events page.
> 
> Any reason that is not a good idea?
> 
> If no objection there; what do you think about omitting "Include" from "Include past events"? If users check 'Past events' it would filter past events only.
> 
> Also, can we anchor 'Past events' at the top of the [Date] list?
> 
> Curious what @valgrimm and @stevejalim think about the above.

I agree about Past Events at top of event list. I also think Past Events should be displayed by default at the bottom of the list IF pagination is already solved. If pagination is not solved, Past Events should not be displayed by default.

valgrimm:
> 
> 
> @valgrimm @malqinneh updated on dev - let me know what you think
> 
> @valgrimm I'm noticing some card images on dev aren't great in terms of fit/flow/sizing now we're using the new card cSS - something to discuss in Slack?

I propose that the new card CSS be pushed to Stage and Stage updated with content from Production, so I can experiment with fixing that problem before the redesign is pushed live. We can discuss, but I want to wait until more development is done.

valgrimm:
Actually, hm, the Past Events filter should work like other filters; when it is checked, only the filtered events are displayed. Currently it does not work that way.
I propose that:
If Past Events is not checked, Past Events will not be displayed.
If ONLY Past Events is checked, ONLY Past Events will be displayed.
If Past Events and another month(s) are checked, Past Events AND events in that month(s) will be displayed.

stevejalim:
@valgrimm Thanks for the feedback and suggestion. Makes good sense. 

Have implemented it and deployed to dev for you to review.

(My hope is we can get this merged into the main events branch today)

malqinneh:
Holding off on feedback via Git comment as per @valgrimm's recommendation. Will schedule a time to meet via Zoom to hash this out.

stevejalim:
@valgrimm @malqinneh Final/confirmed behaviour reinstated (show all by default; show only past if ONLY 'past' is selected; show mix of past and future if that's what selected). Thanks for firming it up. 

Deployed to dev.

Issue #1267, opened 2020-03-17
Design refresh: Event detail page


stevejalim:
AC
- [ ] Looks like design
- [ ] HTML valid

stevejalim:
@malqinneh @valgrimm Before I pick this up, it would be good to know if we are going to continue to show all the details that an event page can show (full address, etc) or if they are being dropped (they're not in the mocks in Abstract). 

eg [here](https://developer-portal.dev.mdn.mozit.cloud/events/perfmatters-conference-2020/) we have an event with venue details.

It's also currenlty possible (in data terms, even if we don't leverage it much in production) to show:
* speakers at the event
* topics associated with the event

I can freestyle support for the following, but if there's resource to get the mockup extended, that'd be great:

* venue details (text)
* venue location (map)
* speakers at the event
* topics associated with the event

malqinneh:
Thank you Steve! Spoke with Val, here are some decisions from that conversation:

- We need to add a secondary button next to [Register] that takes users to the official website of the event page. Button should be labeled as [Official Website]
- We need the ability to add another button that takes you to a specific URL labeled as [Event Content]
- Val agrees to leave out the address due to complications (venue changing, etc.)

Issue #1233, opened 2020-03-11
Design-refresh: page-by-page review/sweep


stevejalim:
Seeking inconsistencies with design and/or between browsers, plus glitches

- [x] HTML Validation re-checks
  - [x] Particularly when SVGs are in play and may need still cruft removing from them

- [x] Cross-browser re-testing
  - [x] Desktop layout
  - [x] Mobile layout
  - [ ] Medium-breakpoint layout

## Snags that need fixing
- [x] People page: topic cards
  - [x] (P1) short title (eg "CSS") is too short to force reflow, needs to be done _with_ CSS --> `width: 100%;` ?
  - [x] (P1) glitched layout in mobile mode -> `body.person .topic-links .mzp-c-card-content` needs to be block, not inline-block (hopefully easy fix)

- [x] (P1) IE11 newsletter signup fails: 'fetch' is undefined.

- [x] (P2) Mozilla Developer logo: Shift 1px or 2px higher compared to menu text?

- [x] (P2) Events list: optional rich-text area/CTA button could do with better positioning

- [x] (P1) Filter list when there's a "Cancel X" link present because fiters are active, needs better spacing

- [x] (P1) IE11: video lightbox not working -> `Object doesn't support property or method "assign"`. Polyfill time.

- [x] (P2) Mobile view of homepage: CTA bar could do with a bit more space above button

- [x] (P2) FILTER LIST Mobile Safari shows rounded-corner button

- [x] Bad hover state on topic cards (see [here](https://app.zenhub.com/workspaces/mdn-developer-portal-5d41d915e27a7b3dba9111ad/issues/mdn/developer-portal/1233#issuecomment-620579790))

## Needs discussion + own tickets

- [x] (P2) Currently the Posts page does not display the header image - do we want to show it? If so, how? --> TICKETED AS #1421

- [ ] (P2) middle-size BREAKPOINT with three/four reflows poorly, compared to mozilla.org -- see how they do it, could be an easy fix? TICKETED AS #1422

## Snags from code review (need own tickets)
- [x] drop use of `section` unless it really really makes sense
- [x] reinstate `role=main` where appropriate
- [ ] add `display: block` to button stacks in mobile + change to UL [see here](https://github.com/mdn/developer-portal/pull/1313#pullrequestreview-383680148)  -> can live without this
- [x] address reflow of header SVG icon on mobile [see here](https://github.com/mdn/developer-portal/pull/1313#issuecomment-605885294)

[...]

PR #1313, opened 2020-03-27
1267: Event detail page design refresh

stevejalim:
This changeset adds a refresh of the Event page.

(Related issue #1267)

## Key changes:

- Data model and Wagtail editing UI updated to hide fields we no longer need and add two new ones:
  - link to 'event content'
  - link to 'official website'

- New design laid in, including
    - header w with Event icon
    - switching to main + sidebar layout
    - new buttons to link out to relevant pages, which appear only when configured
    - sidebar panel shows a card for the same event, and links out to the "most appropriate" thing, which would be the first available link from: event content, then event official site, then registration page. This is better than the card linking to the page it is on.

- Checked on desktop and mobile layouts
- HTML valid 

## How to test

- The code is up on dev and I've made a sample page: https://developer-portal.dev.mdn.mozit.cloud/events/test-event/

(We should do a cross-browser sweep once all the pages are in, but any cross-browser checks you can do are very welcome right now)

## Screenshots

Screenshot_2020-03-27 event x(1)

Screenshot_2020-03-27 event x


schalkneethling:
I wonder if it is worth showing the icon on mobile in the header:

Screenshot 2020-03-30 at 11 21 03

See below the layout without the icon

Screenshot 2020-03-30 at 11 22 25

schalkneethling:
On iPad in portrait mode (768x1024) the buttons and share component alignment seems incorrect:

Screenshot 2020-03-30 at 11 23 43

schalkneethling:
Also on mobile, I wonder if these buttons:

Screenshot 2020-03-30 at 11 25 27

Should not be stack as below?

Screenshot 2020-03-30 at 11 25 52

Setting `display: block;` on this anchor elements will do the trick.

developerportal/apps/events/templates/event.html
```diff
@@ -1,57 +1,73 @@
...
+        {% if page.official_website %}
+          <a href="{{page.register_url}}" class="mzp-c-button">
+            Official Website
+          </a>
+        {% endif %}
+        {% if page.event_content %}
+          <a href="{{page.register_url}}" class="mzp-c-button">
+            Event Content
+          </a>
+        {% endif %}
```
schalkneethling: Seems like this is a candidate for an unordered list?

stevejalim: Good call - will make the change as part of a mop-up ticket, if that's ok. Noted in #1233

valgrimm:
@stevejalim Proposal: On mobile, pop the calendar icon in above the title, to be consistent with the way the icon for  video is handled. Just noticed this, partially much thanks to schalk.

stevejalim:
@valgrimm Actually, I was about to say that I think we should drop it, to avoid making the header unnecessarily tall (ie, go with what Schalk is suggesting). 

(We'd do the same for all mobile headers, then)

Can i steer you the other way? 😄

developerportal/apps/events/templates/event.html
stevejalim: Thanks for the pointer about EOL @schalkneethling! Noted in the mop-up ticket

valgrimm:
@stevejalim For now, yes, drop the icons on mobile. Handling all of the icons responsively will add overhead we cannot afford right now. I'll add a ticket to handle this after the redesign is rolled out. As long as the solution to these icons is handled consistently for all pages, I am good with it. #1323 is the ticket for making these icons responsive @malqinneh This is a short-term solution to reduce scope.

developerportal/apps/events/templates/event.html
stevejalim: Done!

Issue #1316, opened 2020-03-27
Update hero and newsletter images


stevejalim:
AC

- [ ] New images used
- [ ] @1x and @2x supported

PR #1320, opened 2020-03-29
1316 Update hompeage hero and newsletter imagery

stevejalim:
This changeset swaps out the temporary images for the homepage hero and for the newsletter and replaces them with 1x and 2x versions, as appropriate

The hero image position may need some tweaking, but as an initial pass, this is good enough for getting feedback. Might even be fine as-is!

(Related issue #1316)

## How to test

- staged on dev. Cross-browser tested already, but please do try to break the layout!

## Screenshot

Screenshot 2020-03-29 at 23 06 08


developerportal/templates/organisms/newsletter-signup.html
```diff
@@ -3,7 +3,13 @@
 <div class="mzp-l-content">
   <aside id="newsletter" class="newsletter-signup mzp-c-newsletter">
     <div class="mzp-c-newsletter-image">
-      <img src="{% static 'img/newsletter-promo.jpg' %}" alt="Newsletter signup promotional image">
+      <img
+        srcset="
+          {% static 'img/promo/newsletter@1x.jpg' %},
+          {% static 'img/promo/newsletter@2x.jpg' %} 2x"
+        src="{% static 'img/promo/newsletter@2x.jpg' %}"
```
schalkneethling: I am wondering, do we really want to fallback to the 2x asset for browsers that do not understand `srcset`? I believe generally browsers that do not support `srcset` will be older and likely be users with bandwidth-constrained connections.

developerportal/templates/organisms/newsletter-signup.html
```diff
@@ -3,7 +3,13 @@
...
+        alt="Newsletter signup promotional image"
```
schalkneethling: This seems to be purely presentational so `alt=""` will be perfectly fine.

Issue #1261, opened 2020-03-17
Design refresh: People page


stevejalim:
AC
- [ ] Looks like design
- [ ] HTML valid

PR #1327, opened 2020-03-30
1261: People page refresh

stevejalim:
This changeset refreshes the design of the People page.

Note: it shows a standfirst/description if the page is configured with one, despite the design NOT showing one. Showing one is standard behaviour and the flexibility to have something there to explain the page's content is useful.

We also throw away some redundant code here \o/

(Related issue #1261)

## How to test

- On dev

## Screenshots
Screenshot_2020-03-30 People
Screenshot_2020-03-30 People(1)


schalkneethling:
This is looking good. I did pick up this problem in IE11

Screenshot 2020-03-31 at 04 21 45

The icon on the left works when interacted with but, the one on the right does not respond to interactions.

stevejalim:
@schalkneethling Thanks for the spot! That's a known issue in the current design with IE11, but will note it in the snag list for the redesign too

Issue #1336, opened 2020-04-01
Optimise SVG assets for upload and provide to Val


stevejalim:
The new Topic icons will need to be manually uploaded when we deploy the new design.
The assets provided by @malqinneh came from Sketch and needs some markup cruft removing (no problem).
Once this is done, the relevant "cleaned" Topic icons should be made available to @valgrimm 

AC
- [ ] All new SVG icons are cleaned up and still look correct
- [ ] All new SVG icons are placed somewhere stable and accessible to @valgrimm

stevejalim:
[https://app.zenhub.com/files/186898396/bb9d0607-d14d-43f2-a72d-b37b913f6dc1/download](https://app.zenhub.com/files/186898396/bb9d0607-d14d-43f2-a72d-b37b913f6dc1/download)

PR #1403, opened 2020-04-28
1336: Update SVG icons 

stevejalim:
This changeset addresses adds in some improved SVG icons in one of two places in the codebase:

1) For pages where icons are fixed - eg Post (video and article), Event, Person - we now use the latest cut of the icons from @malqinneh. This includes some HTML and CSS tweaks to standardise the positioning and ensure that the icons definitely appear as black-on-white or white-on-black, as appropriate for the colour of their surrounding/background element.

2) For Topic pages, where the icons are uploaded into the CMS, the pre-existing directory of icons has been updated with latest designs, in black-on-white form, ready for upload. _These have already been uploaded to the dev site, for QA_, so all that will remain pre-go-live of the new design is to upload them to the stage and then prod CMSes.

(Related issue #1336)

## How to test

- The code is deployed to Dev, and - as mentioned - the manual-uploadable icons have already all been uploaded to Dev, so a drive around on there is all that's needed.
- Cross browser testing has already taken place.


stevejalim:
@schalkneethling Thank you for the simplified Voice SVG, however for now I am sticking with the original because I noticed it's slightly differently proportioned and I need to check how closely we need to stick to the brand. I can swap it in later, no worries.

data/devportal_topic_icons/people.svg
```diff
@@ -1,9 +1,8 @@
...
-    </g>
-</svg>
\ No newline at end of file
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
stevejalim, line 2: Because the file is only ever loaded and not inlined, I think the straggling doctype here is OK. HTML5 validator is happy

PR #1415, opened 2020-05-01
1233  add ie11 polyfills

stevejalim:
This changeset addresses functional failures in IE11 due to unsupported JS APIs

* Fix newsletter signup failure in IE11 due to lack of native browser support

  * Add polyfill for `fetch` (which in turn required a `Promise` polyfill)
  * Add workaround for absence of URLSearchParams in IE11 and a fully-featured FormDaga implementation - rather than adding yet more (large) polyfills, some more manual code was a more workable solution

* Add in Object.assign polyfill (from MDN docs) so that the light box will work for viewing external videos within the site in IE11

(Related issue #1233)

## How to test

- Code will be staged on the dev server.  Using Browserstack please confirm you can:
* Sign up to the newsletter from the footer in IE11 and also in evergreen Firefox
* View an external video (see the one at the top of the /posts/) page in a lightbox in IE11 and evergreen Firefox

### DIFF
--- a/src/js/newsletter-subscription.js
+++ b/src/js/newsletter-subscription.js
@@ -6,6 +6,7 @@ const NEWSLETTER_SUBSCRIBE_URL = 'https://www.mozilla.org/en-US/newsletter/';
  * @returns {string} URI-encoded string
  */
 function getPayloadString(form) {
+  let retval = null;
   if (typeof URLSearchParams === 'undefined') {
     // if we lack URLSearchParams or a fully-featured FormData (eg: IE11)
     // let's just manually extract the data we need
@@ -16,11 +17,11 @@ function getPayloadString(form) {
       const field = formInputs[i];
       pairs.push(`${field.name}=${encodeURIComponent(field.value)}`);
     }
-    return pairs.join('&');
-    // eslint-disable-next-line no-else-return
+    retval = pairs.join('&');
   } else {
-    return new URLSearchParams(new FormData(form)).toString();
+    retval = new URLSearchParams(new FormData(form)).toString();
   }
+  return retval;
 }
 
 /**
--- a/src/js/newsletter-subscription.js
+++ b/src/js/newsletter-subscription.js
@@ -6,9 +6,7 @@ const NEWSLETTER_SUBSCRIBE_URL = 'https://www.mozilla.org/en-US/newsletter/';
  * @returns {string} URI-encoded string
  */
 function getPayloadString(form) {
-  try {
-    return new URLSearchParams(new FormData(form)).toString();
-  } catch (err) {
+  if (typeof URLSearchParams === 'undefined') {
     // if we lack URLSearchParams or a fully-featured FormData (eg: IE11)
     // let's just manually extract the data we need
     const formInputs = form.getElementsByTagName('input');
@@ -19,6 +17,9 @@ function getPayloadString(form) {
       pairs.push(`${field.name}=${encodeURIComponent(field.value)}`);
     }
     return pairs.join('&');
+    // eslint-disable-next-line no-else-return
+  } else {
+    return new URLSearchParams(new FormData(form)).toString();
   }
 }
 
--- a/src/js/newsletter-subscription.js
+++ b/src/js/newsletter-subscription.js
@@ -1,5 +1,27 @@
 const NEWSLETTER_SUBSCRIBE_URL = 'https://www.mozilla.org/en-US/newsletter/';
 
+/** Opinionated function for extracting data from the form without needing FormData
+ *
+ * @param {object} form element
+ * @returns {string} URI-encoded string
+ */
+function getPayloadString(form) {
+  try {
+    return new URLSearchParams(new FormData(form)).toString();
+  } catch (err) {
+    // if we lack URLSearchParams or a fully-featured FormData (eg: IE11)
+    // let's just manually extract the data we need
+    const formInputs = form.getElementsByTagName('input');
+    const pairs = [];
+
+    for (let i = 0; i < formInputs.length; i += 1) {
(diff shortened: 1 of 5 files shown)


src/js/newsletter-subscription.js
```diff
@@ -1,5 +1,27 @@
...
 
+/** Opinionated function for extracting data from the form without needing FormData
+ *
+ * @param {object} form element
+ * @returns {string} URI-encoded string
+ */
+function getPayloadString(form) {
+  try {
+    return new URLSearchParams(new FormData(form)).toString();
+  } catch (err) {
```
stevejalim, line 11: @Gregoor I think this approach is a bit too 'Pythonic' and it might be better/safer to check if `URLSearchParams` exists, but I can't do that without raising an exception (eg `typeof(URLSearchParams === "undefined")` raises an exception. Any advice?

stevejalim:
@valgrimm Cheers for the eyes!

To test the video-in-a-lightbox, this is the one I was referring to, because it's an External video which will use the popup/lightbox - the HJ Chen on is a video which sits in its own page inside the portal, so has no need to be shown in a lightbox because it already has 'a home' in the site

Screenshot 2020-05-04 at 11 30 53

valgrimm:
Works for me, still approved.

src/js/tests/newsletter-subscription.test.js
```diff
@@ -91,3 +91,60 @@ test('Happy path', async () => {
     '<b>Thank you. Please check your email to confirm your subscription.</b>',
   );
 });
+
+test('IE11 manual fallback', async () => {
+  // Testing that the overall behaviour still works with IE11, which has
```
Gregoor, line 96: micro-nit: I'd move that comment above the test which makes it look more JSonic than Pythonic 🦔 🐍

src/js/tests/newsletter-subscription.test.js
```diff
@@ -91,3 +91,60 @@ test('Happy path', async () => {
...
+  });
+  form.submit();
+
+  expect(fetch.mock.calls.length).toEqual(1);
+
+  expect(fetch.mock.calls[0][1].headers).toEqual({
+    'X-Requested-With': 'XMLHttpRequest',
+    'Content-type': 'application/x-www-form-urlencoded',
+  });
+  expect(fetch.mock.calls[0][1].method).toEqual('POST');
```
Gregoor, line 123: [`.toHaveBeenCalleWith`](https://jestjs.io/docs/en/expect#tohavebeenlastcalledwitharg1-arg2-) reads better imo

src/js/tests/newsletter-subscription.test.js
```diff
@@ -91,3 +91,60 @@ test('Happy path', async () => {
...
 });
+
+test('IE11 manual fallback', async () => {
+  // Testing that the overall behaviour still works with IE11, which has
+  // no URLSearchParams or complete FormData implementation
+  // Store original implementation
+  const originalURLSearchParams = URLSearchParams;
+
+  // eslint-disable-next-line no-global-assign
+  URLSearchParams = jest.fn(() => {
```
Gregoor, line 102: I _think_ this should be safer for mocking globals, as it _shouldn't_ possibly interfere with other test suites
https://stackoverflow.com/a/40460395

stevejalim: So you're saying you agree that it's OK, yeah? (We also re-assign the original global later, too)

src/js/tests/newsletter-subscription.test.js
stevejalim: I'll meet you half way 😄 - see the diff now (in this part and earlier in the file)

src/js/newsletter-subscription.js
Gregoor: Whoops missed that one. You were close, the parantheses lead to a wrong order of execution. It should be: `typeof URLSearchParams === "undefined"`.

Otherwise you get the following evaluation in IE11:
`typeof(URLSearchParams === "undefined")`
`typeof(throwReferenceError('URLSearchParams') === "undefined")`

and in browsers where it exists you'd get
`typeof(URLSearchParams === "undefined")`
`typeof(false)`
`"boolean"`

src/js/tests/newsletter-subscription.test.js
Gregoor: I was trying to recommend a different approach but now that I read more about it I think I was indeed just agreeing and there's no difference in using vs not using the `global` object. So all's well 👍

src/js/newsletter-subscription.js
stevejalim: Ugh, that parenthesis mistake was me fat-fingering things in the IE11 console via Browserstack. Thanks for paying it out so clearly

stevejalim:
@Gregoor Have updated. Could you give it another scan whenever you've got spare time, please? ❤️

src/js/newsletter-subscription.js
```diff
@@ -1,5 +1,28 @@
...
+    // let's just manually extract the data we need
+    const formInputs = form.getElementsByTagName('input');
+    const pairs = [];
+
+    for (let i = 0; i < formInputs.length; i += 1) {
+      const field = formInputs[i];
+      pairs.push(`${field.name}=${encodeURIComponent(field.value)}`);
+    }
+    return pairs.join('&');
+    // eslint-disable-next-line no-else-return
```
Gregoor, line 20: Not sure if I'm a fan of the lint rule, but in this context it might be easier to just follow it, move the return outside of else and drop the branch

stevejalim: Noted, and yeah I agree. Fixed!

PR #1426, opened 2020-05-04
1264: Simple styling for the Videos page (at /videos/) just in case i…

stevejalim:
This changeset adds simple/basic styling for the Videos page (at /videos/) just in case it's ever published in error. It is not a highly functional page, just a list of links to all published `Video` pages in the CMS.

**Context:** 
The Videos page has to exist in the CMS ( at /videos/) so that it can be the parent page of actual Video pages eg (/videos/my-video-page/), but it is not intended to be shown in menus nor accessed directly, so it is only ever in existence in draft form.

As such, this template only has basic styling as a simple safety net in case the page is made live (either in error or deliberately, before additional styling is added)

## How to test

- Code is up on Dev. View the Draft of the Videos page to see what the page WOULD look like if it were ever published: https://developer-portal.dev.mdn.mozit.cloud/admin/pages/5/view_draft/

Example from local dev:
Screenshot 2020-05-04 at 12 46 28

### DIFF
--- a/developerportal/apps/videos/models.py
+++ b/developerportal/apps/videos/models.py
@@ -49,6 +49,11 @@ class VideoTopic(Orderable):
 
 
 class Videos(BasePage):
+    # IMPORTANT: The Videos page has to exist in the CMS ( at /videos/) so that
+    # it can be the parent page of actual Video pages eg, but it is not intended
+    # to be shown in menus nor accessed directly, so it is only ever
+    # in existence in draft form.
+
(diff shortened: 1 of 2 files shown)


stevejalim:
@valgrimm 
> Thought it would have cards like Posts, 

Yeah, originally I thought it'd look like that, too, but then I balanced up the time/effort for a page that will remain hidden and went with the MVP solution

> but a list of links works too, as this isn't a page for publication. I'll just say "looks good to me"

Thanks!

Issue #697, opened 2019-11-05
Run test against kubernetes manifest using kubeval


limed:
Create a `k8s-test-manifest` target against everything in the Makefile that requires an apply before we run a deployment.

See: https://kubeval.instrumenta.dev/

limed:
This is done

PR #1445, opened 2020-05-07
Added a test target to test k8s manifest

limed:
Test kubernetest manifest before deploying, utilizes kubeval which has been installed on the CI server

(Related issue #697)

## Key changes:

- Added tests for deployment to test if k8s manifest is valid

## How to test

- Install [kubeval](https://github.com/instrumenta/kubeval) on your local machine 
- Source config file
```
cd k8s
source k8s/config/${dev,stage,prod}.sh
```
- Run make file to run test
```
cd k8s
make test-k8s-deployments
```

### DIFF
--- a/k8s/Makefile
+++ b/k8s/Makefile
@@ -94,10 +94,15 @@ k8s-delete-ns:
 k8s-services:
 	j2 app.svc.yaml.j2 | ${KC} apply -f -
 
+test-k8s-services:
+	j2 app.svc.yaml.j2 | kubeval --strict --ignore-missing-schemas
+
 k8s-delete-services:
 	${KC} delete --ignore-not-found svc ${APP_SERVICE_NAME}
 
 k8s-deployments: k8s-wagtail-deployments k8s-celery-deployments k8s-celery-beat-deployments
+
+test-k8s-deployments: test-k8s-wagtail-deployments test-k8s-celery-deployments test-k8s-celery-beat-deployments
 
 k8s-delete-deployments: k8s-delete-wagtail-deployments k8s-delete-celery-deployments k8s-delete-celery-beat-deployments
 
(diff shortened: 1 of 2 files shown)


k8s/Makefile
```diff
@@ -94,29 +94,50 @@ k8s-delete-ns:
 k8s-services:
 	j2 app.svc.yaml.j2 | ${KC} apply -f -
 
+test-k8s-services:
+	j2 app.svc.yaml.j2 | kubeval --strict --ignore-missing-schemas
+
```
stevejalim, line 99: Is this addition separate from what `test-k8s-deployments` uses? Should it be included in there, too?

k8s/Makefile
```diff
@@ -94,29 +94,50 @@ k8s-delete-ns:
...
+	env NEW_RELIC_APP_NAME=dev-portal-celery-${TARGET_ENVIRONMENT} \
+	env DJANGO_SETTINGS_MODULE=developerportal.settings.worker \
+	j2 celery.yaml.j2 | kubeval --strict --ignore-missing-schemas
+
 k8s-celery-beat-deployments:
 	env NEW_RELIC_APP_NAME="" \
 	j2 celery-beat.yaml.j2 | ${KC} apply -f -
 
+test-k8s-celery-beat-deployments:
+	env NEW_RELIC_APP_NAME="" \
```
stevejalim, line 132: Should that name be populated? Or is it empty because we're not checking on the status of celery beat in New Relic?

k8s/Makefile
```diff
@@ -94,29 +94,50 @@ k8s-delete-ns:
...
+	j2 celery-beat.yaml.j2 | kubeval --strict --ignore-missing-schemas
+
 k8s-hpa:
 	j2 hpa.yaml.j2 | ${KC} apply -f -
 
+test-k8s-hpa:
+	j2 hpa.yaml.j2 | kubeval --strict --ignore-missing-schemas
+
```
stevejalim, line 140: Another additional command but not part of the one called for Slack, yeah?

k8s/Makefile
limed: Yeah it shouldn't be populated because we don't really care about it

limed:
> Heya - tried this locally but couldn't get it to run.
This might be because you need the kubeval binary? If you have an error message that would be cool

k8s/Makefile
limed: Apologies not quite following this here :)

Issue #1438, opened 2020-05-06
Design Refresh: Support a 3:2 aspect ratio image for the card_image


stevejalim:
Context:

The redesign has introduced the idea of 16:9 aspect-ratio images, not just 3:2 as the original design had. This means we can end up with the same image being cropped in two ways, depending on which card is is shown in.

Also, at the moment, even though we have `Page.image` and `Page.card_image,` we're only really using `Page.image`, and this will soon be renamed to `Page.card_image`, because for semantic reasons, `Page.image` should be deprecated because it is no longer rendered in the page, only in the Card.

So, taking those two things into account, this piece of work needs to meet meet the following AC:

- [ ] Each Page's `Image` field is removed, and whatever image it referenced is now set as the value of the Card tab's `Image` field. 
- [ ] The help text for this field makes it clear that the image is expected to be 16:9 in aspect ratio. 
- [ ] The Card tab also has a new `3:2 Image` field, for a separate image, with help text that clarifies the required aspect ratio for this field. 
- [ ] The UI code will NOT enforce the aspect ratio when selecting any image
- [ ] In the site, when a card is shown (eg on a Topic page that has two, four or five featured items), any cards that should be 3:2 aspect ratio will seek to use that new 3:2-ratio card image. 
- [ ] If one is not configured, the default behaviour will be to use in the 16:9 image, automatically cropped as per standard wagtail behaviour (for now, pending behaviour of separate black-bar-image-letterboxing ticket)

PR #1455, opened 2020-05-11
1438: Add new aspect ratio card image

stevejalim:
This changeset follows the clean-up work in PR #1454  and adds support for a new "card image", this this time in a 3:2 aspect ratio rather than the new default of 16:9.

If the 3:2 image is not available/not configured, a 3:2 fallback will be used.

(Related issue #1438 )

## Key changes:

-  Add support for a 3:2-aspect-ratio card image on certain Page types

    Specifically those which can appear as 'Featured' items, where certain configurations can result in 3:2 aspect-ratio images being used instead of the new default of 16:9:
    * `Article`s
    * `Video`s
    * `Event`s
    * `ExternalContent` (ie, `ExternalArticle`s, `ExternalVideo`s, `ExternalEvent`s)

    Note that `ContentPage`, `Topic` and `Person` Page types do NOT get the 3:2 field added, because we don't need it yet and it's confusing to add it if we don't use it

-  Add support for a 3:2 image in all our custom StreamBlocks
    The 3:2 image is optional, just to make life slightly easier for cases where we know we don't need to worry about it being rendered as a 3:2 image.

- Complete support of 3:2 images for featured items by picking the appropriate image in the HTML

    Includes:
    * Pages (ie, using the appropriate "card image")
    * StreamBlocks (ie, using the appropriate "image" field)
    * Fallback images if either of the above is needed but not present

-  FIX: Ensure the bottom row of featured items in a two-rows-of-two layout uses a 3:2 aspect ratio image, too

## Screenshots

**Example of a page now featuring both 16:9 and 3:2 image options**
Screenshot 2020-05-11 at 11 52 21

**Example of StreamField block (ie ad-hoc content, not a page) now featuring both 16:9 and 3:2 image options** -- this is useful for when we want to feature an ad-hoc external link on the homepage, for instance - but not that the importer will not automatically import a 3:2 image, so that will need manual handling as/when required.

Screenshot 2020-05-11 at 13 08 44

**Example of a featured external link switching between 3:2 and 16:9 images** depending on where it is configured. (ie, this is a `StreamField`'s `StreamBlock` showing the behaviour, but a featured Page will also behave the same way.
Screenshot 2020-05-11 at 13 13 22
Screenshot 2020-05-11 at 13 13 43

**And an example of a 3:2 card lacking a 3:2 image falling back to the correct-aspect-ratio placeholder**
Screenshot 2020-05-11 at 13 22 04

## How to test

- Code is up on dev. I've also added a test 3:2 and test 16:9 image. Please drop these into various pages and featured-elements configurations (including re-ordering them) and previewing/publishing the pages to see different behaviour. Please try to break it!


valgrimm:
How do I activate the 3:2 image? Shrinking my browser window doesn't seem to do it.

stevejalim:
@valgrimm 

> How do I activate the 3:2 image? Shrinking my browser window doesn't seem to do it.

Ah, apologies. This isn't a dynamic swapping of a 3:2 image for a 16:9 based on browser width - it's just about using a 3:2 image rather than the 16:9 version when there are two images in a row of featured items. So, to test out the behavoiour, you should try re-ordering the items in a Featured block on the homepage or a Topic page (you can use the `^ v` arrows in the editor UI for the Featured item, or drag and drop them around)

Hope that helps!

stevejalim:
@Gregoor 

> Wondering if there's a way to share more of the feature page type's code, given how much we have to repeat ourselves for changes like this.

Fair comment! We could move some of the repeated code into mixins, but then you walk the line of class-based/inheritance soup, I think. I'll note it as a possible refactor/cleanup later, though

Issue #1382, opened 2020-04-14
Add breathing room to H2 spacing


valgrimm:
AC: Add 1 line of spacing above H2 titles to match design

(otherwise this has to be added manually with a < br >, as has been done on the Hubs page)

stevejalim:
Can you add a screenshot of what you want it to look like and on which page, please? (ie am assuming you're talking about a content page/rich text, but need to check as we have a lot of other H2s in the site)

valgrimm:
The top image below shows a page where I have added two brs between the image and the heading. The bottom image below shows one where I have removed those brs.

Spacing should be as in the top image by default. 

![withbrbr](https://user-images.githubusercontent.com/56933147/81583121-2ce63400-93b1-11ea-9ffa-95652227738e.PNG)
![withoutbr](https://user-images.githubusercontent.com/56933147/81583126-2f488e00-93b1-11ea-8f7a-7ad285538272.PNG)

stevejalim:
@valgrimm Rather than mandate a H2 has to have space above it, which would mean that a content starting with a H2 would always have a space at the top, I'll add a bottom margin to the image, resulting in the same effect

valgrimm:
Sounds like a good solution to me (:

PR #1474, opened 2020-05-14
1382: Improve spacing after elements in a content page

stevejalim:
This changeset fixes up the natural spacing on a `ContentPage` so that no hacks with `<br>` in HTML blocks are needed.

It puts a 48px margin below certain content blocks that are allowed in the `CustomStreamField` in the `ContentPage`

Blocks given spacing after them are: Paragraph, Image, Embed or Code snippet

The raw HTML block does NOT get automatic spacing - that seemed unnecessary

Note that it does NOT do this for `CustomStreamField` use anywhere else (eg on an Event page or similar) - that's something we can review as needed

(Related issue #1382)

## Screenshots

**BEFORE**
Screenshot_2020-05-14 Test topic subpage (For CSS in this case)(1)

**AFTER**
Screenshot_2020-05-14 Test topic subpage (For CSS in this case)(3)

## How to test

- Code is staged on dev. I had edited [this page](https://developer-portal.dev.mdn.mozit.cloud/admin/pages/379) to remove the HTML `br` blocks and to also put the H2 in the SAME block as the text which follows it (else the H2 will have a bottom margin too). Please confirm you're happy with both the spacing and the approach to authoring the page (it's now less fragmented, which should be less painful!)

### DIFF
--- a/src/css/templates/content-page.scss
+++ b/src/css/templates/content-page.scss
@@ -6,6 +6,12 @@ body.content {
   div.syntax-highlight {
     margin-bottom: $spacing-2xl;
   }
+
+  .standfirst-wrapper {
+    .rich-text {
+      margin-bottom: 0px;
+    }
+  }
 }
 
 .mzp-has-sidebar {
--- a/src/css/templates/content-page.scss
+++ b/src/css/templates/content-page.scss
@@ -1,3 +1,13 @@
+body.content {
+  figure,
+  div.rich-text,
+  p.button-block,
+  div.responsive-object,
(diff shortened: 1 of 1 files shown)


valgrimm:
The example page seems to be dead, but the screenshot looks good.

stevejalim:
@valgrimm Ugh i left off the trailing slash in the link. Try [this](https://developer-portal.dev.mdn.mozit.cloud/admin/pages/379/edit/)

valgrimm:
approved!

PR #1508, opened 2020-05-27
1488: Update dev-setup to include more realistic data

stevejalim:
This changeset does three things, all related to the developer set-up process:

1) It adds a new fixture based on near-production-level data, minus a few elements that are not easily viable to include in data fixtures. This data is based on the contents of staging shortly after the new design was rolled out.

Images are not included, so all pages will fall back to using their defaults. This includes topic icons. To add topic icons, upload them manually into the Topic pages in the CMS, sourcing them from the `data/devportal_topic_icons` directory.

Note that user data is not loaded either, so to access the CMS and Django Admins, you will have to create a superuser using your mozilla.com email address, or disable SSO for local builds (see `local.py.example`)

2) It updates the `scripts/dev-setup.sh` bootstrapping script has been updated to load this newer, larger dataset by default, but it is also possible (via the `--minimal-data` flag) to load only the original skeleton data structures from common.json (which is still used for unit tests)

Example usage:

`./scripts/dev-setup --prune` - uses `fixtures/full_site_no_images.json`

`./scripts/dev-setup --prune --minimal-data` - uses `fixtures/common.json`

3) It drops the automatic generation of a local database password, in favour of a static one. (This avoids an issue where `dev-setup.sh` generates a new DB password but the actual DB container only knows the old one, so has to be purged in order to work again...) Given that this password will NEVER be used in a deployed site and local machines should be kept secure, plus there is no PII stored in the local database

(Related issue #1488)

## How to test

- check out the branch and run `./scripts/dev-setup --prune` then `docker-compose up` You should see a well-populated local build using placeholder images and icons


stevejalim:
Adding @Gregoor for review, but also @schalkneethling for awareness, because this will help with local setup.

@Gregoor What do you think, in particular, about the changes that
a) get rid of the random local-developement-db password
and
b) use a dump of production-like content (featuring real Mozilla contributors in the People section, etc) as local dev data. I could anonymise them all, of course, but given the site is public and it's all open, how do you feel about it?

Issue #1513, opened 2020-05-28
Posts-page search: add support for searching as part of refining options


stevejalim:
Now the relevant models are being indexed, now we can lay in the backend code that will apply search params to the logic that gets and filters pages for display 

### Acceptance Criteria
 
- [x] Adding a `search=term` querystring to the `/posts/` URL will trigger a search for `term`
- [x] Any Topic filters for the Posts page should be respected if selected, so when combined with search, users are able to further narrow the scope of results. 
- [x] Search results are represented in the same way as standard filtering: whatever results match the overall parameters (search text + Topic filters) are shown as cards in the standard paginated list on the Posts page
- [x] If no results match the given parameters (Search with and without Topic filters) the Posts page results list shows an appropriate message (just the same as if someone had "over-filtered" without specific search terms)

stevejalim:
Bumping up against some unexpected challenges with the tests: I think the test fixtures am creating aren't _quite_ correct, meaning the tests fail but the local build works. Will keep on exploring

Issue #1473, opened 2020-05-14
Posts page: implement backend search behaviour


stevejalim:
## Acceptance Criteria

- [x] A visitor site should be able to search the Title and Description fields for Posts, Videos, External Posts, and External videos - this functionality will be surfaced in #1451 but for now a URL an be amended to trigger search
- [x] Existing Topic filters for the Posts page should be respected if selected, narrowing the scope of results 
- [x] Search results are represented in the same way as standard filtering: whatever results match the overall parameters (search text + Topic filters) are shown as cards in the standard paginated list on the Posts page
- [x] If no results match the given parameters (Search with and without Topic filters) the Posts page results list shows an appropriate message (just the same as if someone had "over-filtered" without specific search terms)
- [x] Unit tests are written that prove this behaviour is correct and protect against regressions in the future

### Dev-notes / Task breakdown 
* Support indexing all Posts (`Article`, `Video`, `ExternalArticle` and `ExternalVideo`) by their `title` and `description` fields (1.5 days)
* If a search string is submitted as part of the filtering form, this should be used to scope-down the queryset _after filtering_  (see [Wagtail Docs](https://docs.wagtail.io/en/v2.2.2/topics/search/searching.html#searching-pages)) (1.5 days)

Issue #1451, opened 2020-05-11
A search field should be present on the Posts page


valgrimm:
## Acceptance criteria

- [x] The filter list (on the left side of a list of Posts) contains a search field below the existing filters and within the bounds of the existing panel
- [x] The search field accepts a single line of text
- [x] The search field features an 'X' at its far right, which will clear the input when clicked
- [x] Hitting "Refine results" triggers the search parameters in concert with any other filters selected in the filtering list
- [x] HTML markup included in search field (eg `<h1>` or `<script>`) is rendered appropriately in the input after a search has been submitted, but escaped appropriately while being processed

PR #1520, opened 2020-06-01
1513: Add backend logic for "Posts"-page search

stevejalim:
This changeset adds backend-only support for searching the `title` and `description` fields of items which appear on the Posts (`/posts/`) page. (Specifically, those are `Article`s, `Video`s, `ExternalArticle`s and `ExternalVideo`s)

This will be followed by #1451 which will provide frontend access to this search behaviour. 

For now, one can test the search via URL manipulation (see below).

Notes:

* While the portal operates behind a CDN, the benefits of that are reduced substantially when search is involved -- cacheing paginated results in `/posts/` is at least potentially relevant to a greater proportion of visitors than cacheing results based on arbitrary search parameters. We will need to monitor load on the k8s cluster and adjust scaling rules appropriately. (Or we rebuild the Posts, Events and People pages entirely, in a worst case - eg to directly query an ES search backend from in-browser JS)

* Deep down, Wagtail's `postgres_search` backend does parameter quoting and escaping (see  See [Lexeme().as_sql()](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/postgres_search/query.py)), but opinions on whether what's currently implemented needs additional protections is very welcome.

(Related issue #1513 and parent #1473)

## How to test

- The code has been deployed to [stage](https://developer-portal.stage.mdn.mozit.cloud/)
- You can trigger a search by adding `search=some%20%terms` as a querystring. eg:
  - https://developer-portal.stage.mdn.mozit.cloud/posts/?search=css%20grid -- try this one in Safari too, to confirm the space works OK; try using pagination if it's offered
  - https://developer-portal.stage.mdn.mozit.cloud/posts/?search=covid-19

- It can also be mixed with existing filter selections from the sidebar:
  - https://developer-portal.stage.mdn.mozit.cloud/posts/?topic=css&search=flexbox 
  but please note that selecting a new topic and hitting "refine results" will lose any search params you have entered into the URL - that will be tackled in #1451

- Please confirm that pagination still works when you have search params in play (eg https://developer-portal.stage.mdn.mozit.cloud/posts/?search=firefox)
 
- We should leave the 'try to break it via the frontend' until #1451, but please do give the backend a grilling


developerportal/apps/common/utils.py
```diff
@@ -17,15 +18,35 @@ def _resolve_model(model):
     return apps.get_model(model) if isinstance(model, str) else model
 
 
+def _prep_search_terms(terms: str) -> str:
+    """Fix up provided terms ready for passing to Wagtail's .search() model
+    manager method. Note that at a lower level, for safety, the postgres_search
+    backend does escaping and quoting for us: See Lexeme().as_sql() in
+    https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/postgres_search/query.py  # noqa
+    """
+    terms = unquote(terms)
```
stevejalim, line 27: The `unquote` step here now seems redundnat, because Django's `request.GET.get` seems to do the legwork anyway. I left it in as a conversation-starter, though, as this is an area that's worth a chat in general

developerportal/templatetags/app_tags.py
```diff
@@ -124,22 +126,27 @@ def pagination_additional_filter_params(request):
...
     for input_key in input_keys:
         if input_key not in QUERY_PARAMS_TO_PASS_ON:
             continue
         input_params = request.GET.getlist(input_key)
         for input_param in input_params:
-            output_params_strings.append(f"{input_key}={input_param}")
+            # Note that using `urllib.parse.quote` avoids spaces appearing in
+            # the output, which breaks Safari (because the space triggers a web
+            # search, not a HTTP request to the site)
+            output_params_strings.append(f"{input_key}={quote(input_param)}")
```
stevejalim, line 149: I went with `quote` rather than `quote_plus`, but can't really pick a firm reason to stand on this hill. What do you think?

developerportal/apps/common/tests/test_utils.py
```diff
@@ -1,11 +1,23 @@
 import datetime
```
stevejalim, line 1: A lot of the code in here is just moved from the redundant and just-deleted `test_common.py`. I'll mark the bits which are new

developerportal/apps/common/tests/test_utils.py
```diff
@@ -49,3 +61,287 @@ def test_paginate_resources__empty_string(self):
         resources = paginate_resources(resources, per_page=10, page_ref="")
         self.assertEqual(repr(resources), "<Page 1 of 3>")
         self.assertEqual([x for x in resources], [x for x in range(1, 11)])
+
```
stevejalim, line 64: This is new

developerportal/apps/common/tests/test_utils.py
```diff
@@ -49,3 +61,287 @@ def test_paginate_resources__empty_string(self):
...
+        )
+
+        for case in cases:
+            with self.subTest(msg=case["desc"]):
+                self.assertEqual(_prep_search_terms(case["input"]), case["expected"])
+
+
+class HelperFunctionTestsWithFixtures(TestCase):
+    fixtures = ["common_plus_extras_for_search_tests.json"]
+
```
stevejalim, line 114: This class has been moved and updated/extended with the final test

developerportal/apps/common/tests/test_utils.py
```diff
@@ -49,3 +61,287 @@ def test_paginate_resources__empty_string(self):
...
+        for _callable in [
+            get_combined_articles,
+            get_combined_articles_and_videos,
+            get_combined_events,
+            get_combined_videos,
+        ]:
+            self.assertEqual(len(_callable(self.page)), 0)
+
+
+class BasePageFormTestCase(TestCase):
```
stevejalim, line 154: This class is old/previously reviewed and unaltered in this changeset

developerportal/apps/common/tests/test_utils.py
```diff
@@ -49,3 +61,287 @@ def test_paginate_resources__empty_string(self):
...
+                form.fields["slug"].help_text,
+                (
+                    "Because you are adding External content, "
+                    "this slug will NOT be visible to the end user, "
+                    "but still needs to be unique within the CMS."
+                ),
+            )
+
+
+class CustomSearchTests(TestCase):
```
stevejalim, line 223: This class is new

developerportal/apps/externalcontent/models.py
```diff
@@ -201,8 +201,10 @@ class Meta:
 
     # Search config
     search_fields = BasePage.search_fields + [  # Inherit search_fields from Page
-        index.SearchField("description")
         # "title" is already specced in BasePage
+        index.SearchField("description"),
+        # Add FilterFields for things we may be filtering on (eg topics)
+        index.FilterField("slug"),
```
stevejalim, line 207: This is a requirement of `wagtailsearch`, because the queryset involved has had a `filter()` performed on it

developerportal/templatetags/app_tags.py
```diff
@@ -124,22 +126,27 @@ def pagination_additional_filter_params(request):
...
+
     for input_key in input_keys:
         if input_key not in QUERY_PARAMS_TO_PASS_ON:
             continue
         input_params = request.GET.getlist(input_key)
         for input_param in input_params:
-            output_params_strings.append(f"{input_key}={input_param}")
+            # Note that using `urllib.parse.quote` avoids spaces appearing in
+            # the output, which breaks Safari (because the space triggers a web
+            # search, not a HTTP request to the site)
```
stevejalim, line 148: This was fun to discover:
Screenshot 2020-06-01 at 14 40 44

developerportal/apps/common/utils.py
stevejalim: Actually, in the next PR, am swapping it for `bleach()` - for reasons I'll explain there

developerportal/templatetags/app_tags.py
stevejalim: Note: maybe it doesn't make a huge difference, as it'll be mitigated in the FE code in the next PR, because Firefox plus-quotes anyway

stevejalim:
@Gregoor Do you have 10-15 mins to check out this one, so I can amend if necessary then merge, please? Cheers!

developerportal/apps/common/tests/test_utils.py
```diff
@@ -49,3 +61,287 @@ def test_paginate_resources__empty_string(self):
...
+            {
+                "desc": "Unescape urlencoded params",
+                "input": "Hello%20World%21",
+                "expected": "Hello World!",
+            },
+            {"desc": "Just spaces", "input": "     ", "expected": ""},
+            {
+                "desc": "Markup allowed (at this level) - handled deeper down",
+                "input": "<script>alert('boo');</script>",
+                "expected": "<script>alert('boo');</script>",
```
stevejalim, line 77: This is amended in the next PR (the FE work for Posts Search)

developerportal/templatetags/app_tags.py
Gregoor: I'm erring towards quote_plus, but don't have a good reason for it yet (I like reading those kinds of URLs better. Google does it.. I guess those are semi-okay reasons).

stevejalim: Cheers. Going to go with `quote_plus` because it matches the behaviour of `x-www-form-urlencoded`, which is what is going on here (it's a form submission, after all)

PR #1547, opened 2020-06-08
1451: Frontend for Posts search

stevejalim:
This changeset adds an updated UI for the Posts page, so that search parameters can be added via the filter form:

Screenshot 2020-06-08 at 13 07 00

Screenshot 2020-06-08 at 12 56 16
Screenshot 2020-06-08 at 13 06 42
Screenshot 2020-06-08 at 13 36 56
Screenshot 2020-06-08 at 13 41 44

(Related issue #1451)

## Key changes:

- search input added, with appropriate sizing for desktop and mobile breakpoints (inlcuding Safari vs mobile Chrome and FF)
- JS updated to handle clearing the search input, showing the "clear" trigger appropriately, etc  
- backend code updated slightly to further sanitise input string - see inline comment in PR 
- cross-browser tested

## How to test

- Code is staged on [Staging](https://developer-portal.stage.mdn.mozit.cloud/posts/) so can be tested without a local build
- Please try any/all search params you might normally try, as well as injection attacks, bad input etc - try to break it!
- confirm pagination works ok in concert with search and filters (eg search for 'CSS' while also having the 'CSS' filter enabled and confirm the URLs for each page are consistent and correct
- confirm the filtering on /events/ and /communities/posts/ still works as normal
- feel free to cross-browser test on real devices, too!


developerportal/apps/common/tests/test_utils.py
```diff
@@ -72,9 +72,9 @@ def test_prep_search_terms(self):
             },
             {"desc": "Just spaces", "input": "     ", "expected": ""},
             {
-                "desc": "Markup allowed (at this level) - handled deeper down",
+                "desc": "Markup escaped via bleach",
                 "input": "<script>alert('boo');</script>",
-                "expected": "<script>alert('boo');</script>",
+                "expected": "&lt;script&gt;alert('boo');&lt;/script&gt;",
```
stevejalim, line 77: This test changed because i introduced `bleach.clean()` to escape markup

developerportal/apps/common/utils.py
```diff
@@ -21,12 +23,14 @@ def _resolve_model(model):
...
     manager method. Note that at a lower level, for safety, the postgres_search
-    backend does escaping and quoting for us: See Lexeme().as_sql() in
+    backend ALSO does escaping and quoting for us: See Lexeme().as_sql() in
     https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/postgres_search/query.py  # noqa
     """
+
     terms = unquote(terms)
     NON_SPACE_WHITESPACE = "\n\t\r"
     terms = terms.translate(str.maketrans("", "", NON_SPACE_WHITESPACE))
+    terms = bleach.clean(terms)
```
stevejalim, line 33: This may seem a bit surprising, to be bleaching on the way in but it serves two purposes

1) I feel better about treating all input as dangerous, even though Wagtail Search also escapes at a lower level

2) Testing locally, with the string `<marquee>` in some test content, I tried searching and - before bleaching to escape the content - a search for "<marquee>" would return no results, even though it should have found the page with the test content. When it the input was bleached, however, it WAS found. (My assumption here is that the sanitised/escaped input better matches the string in the wagtailsearch index, and talking with one of the core team it appears this could well be the case, and is a bug that they will address)

src/css/organisms/filter-form.scss
```diff
@@ -56,6 +56,28 @@
...
+  }
+
+  .filter-form-search {
+    padding: $spacing-xs;
+
+    @media #{$mq-md} {
+      width: 100%;
+    }
+    width: 97%;
+    @supports (-webkit-touch-callout: none) {
```
stevejalim, line 74: This was my way to target iOS Safari uniquely. What do you think?

src/js/organisms/filter-form.js
```diff
@@ -1,4 +1,4 @@
-const { parseForm, parseQueryParams } = require('../utils');
```
stevejalim, line 1: This file might be better reviewed by viewing just the current state, not the diff

src/js/tests/filter-form.test.js
```diff
@@ -0,0 +1,411 @@
+const FilterForm = require('../organisms/filter-form');
```
stevejalim, line 1: This is all new, including backfilling tests for existing code

src/js/utils.js
```diff
@@ -24,20 +23,14 @@ exports.parseQueryParams = () => {
...
+exports.decodeFormURLEncodedSpaces = (value) => {
+  /* Turn "foo+bar" into "foo bar".
+   * Context: decodeURIComponent (used above in parseQueryParams) does not
+   * convert `+` to spaces, because the `+` is from the spec for
+   * application/x-www-form-urlenconded, not encodeURIComponent)
+   *
+   * @param {string} value
+   * @returns {string}
+   */
+  return value.replace(/\+/g, ' ');
```
stevejalim, line 35: Wrapping this in a func is heavy, i know, but i wanted to make it explicit, testable and easily extensible

stevejalim: Equally, it's a minor pain to have to be doing this, but I think it's unavoidable. Some browsers (eg FF) use `+` for spaces, so we have to expect it

developerportal/apps/common/utils.py
stevejalim: Yeah, I thikn I'll add a test that fails and prints a reminder message when the Wagtail version increases - see https://github.com/wagtail/wagtail/issues/6131

- [ ] add canary test

developerportal/templates/molecules/filter-form.html
```diff
@@ -105,6 +105,21 @@ <h5>Country</h5>
...
+      <header class="filter-form-section-header">
+        <h5>Search</h5>
+          <a href="#" class="filter-form-clear js-filter-clear" data-controls="{{SEARCH_QUERYSTRING_KEY}}">
+          <span class="icon">
+            {% include "atoms/icons/close.svg" %}
+          </span>
+          Clear
+        </a>
+      </header>
+      <input class="filter-form-search js-search-input" name="search" value="" placeholder="Refine with keywords"/>
```
schalkneethling, line 119: This is missing an associated `label` element. Not sure if you have the utility class `visually-hidden`, but that can be used on the label to hide it from everyone except screen reader users.

```
.visually-hidden {
    border: 0;
    clip: rect(0 0 0 0);
    height: 1px;
    margin: -1px;
    overflow: hidden;
    padding: 0;
    position: absolute !important; /* stylelint-disable-line declaration-no-important */
    width: 1px;
}
```

developerportal/templates/molecules/filter-form.html
```diff
@@ -105,6 +105,21 @@ <h5>Country</h5>
...
+  {% if filters.show_search_input %}
+    <fieldset class="filter-form-section">
+      <header class="filter-form-section-header">
+        <h5>Search</h5>
+          <a href="#" class="filter-form-clear js-filter-clear" data-controls="{{SEARCH_QUERYSTRING_KEY}}">
+          <span class="icon">
+            {% include "atoms/icons/close.svg" %}
+          </span>
+          Clear
+        </a>
```
schalkneethling, line 117: This should be a `button type="button"` element - Golden rule: "Links take you somewhere, buttons do things" 😸

src/css/organisms/filter-form.scss
```diff
@@ -56,6 +56,28 @@
       padding-bottom: 0px;
     }
   }
+
+  /* - Internet Explorer 10–11
+   - Internet Explorer Mobile 10-11 */
+  :-ms-input-placeholder {
+    font-size: 0.8rem;
+    padding: 4px;
+  }
```
schalkneethling, line 65: I reckon we can gracefully degrade here. Support for `rem` units in those versions is also not great, although it should work with `font-size`.

src/css/organisms/filter-form.scss
schalkneethling: CSS feature detection FTW 😸

src/js/organisms/filter-form.js
```diff
@@ -29,43 +29,76 @@ module.exports = class FilterForm {
     this.state = parseQueryParams();
 
     // Elements for the filter form/list.
+    // IMPORTANT: there are TWO forms in any page featuring this component: one for
+    // desktop viewports and one for mobile/collapsed-menu filtering
```
schalkneethling, line 33: ouch! Do we really need two? It might be a necessary evil, but I thought I would ask.

developerportal/templates/molecules/filter-form.html
stevejalim: Yep, indeed! My lame excuse is I inherited this markup / behaviour, but I guess it's on me now 😄

src/js/organisms/filter-form.js
stevejalim: Again, an inherited setup, which could probably be addressed be refactoring it all with flexbox and flex-direction, but it's not been a big priority (and I doubt we have scope right now, but maybe in a couple of months)

developerportal/templates/molecules/filter-form.html
stevejalim: Because it's a pre-existing problem, am going to split this point out into a separate ticket (but it won't get lost, I promise) so I can ship this larger piece sooner. Hope that's ok with you

stevejalim: Ticketed as #1564 1564

Issue #1532, opened 2020-06-03
Event page: Implement backend search behavior


valgrimm:

valgrimm:
Same AC as #1473, but for Events page

PR #1553, opened 2020-06-09
1532: Amend Events page date filtering to be broader, with sensible default

stevejalim:
This changeset removes the behaviour of showing all future events by default on the Events page and instead limits the default view to ones in the next six calendar months. This is so that more relevant, sooner-to-happen Events appear on the page, without far-future Events appearing before them in the results.

It also removes the option to filter by future month and replaces it with a 'future events' checkbox filter, to complement the existing 'past events' filter.

If neither Future Events nor Past Events is selected, the user sees any events happening within the next six months.
If only Future Events is selected, the user sees only future events
If only Past Events is selected, the user sees only past events
If both are selected, the user sees all events in the CMS

(All of the above are still paginated, naturally)

Note that if there are no events happening in the next six months, the default view will show "No relevant events found", so the CMS must be kept up to date with applicable events.

The filter-form partial shows a special message when the default six-month-windowing is in use.

**Please note the ordering in these screenshots is not the current/final ordering: it has been updated to show "least-far-in-the-future first" (ie, regular chronological ordering) - please see discussion below**

Screenshot 2020-06-09 at 16 55 28
----
Screenshot 2020-06-09 at 16 55 55
----
Screenshot 2020-06-09 at 16 59 07
----

(Related issue #1532)

## How to test

- Code is staged on [Staging](https://developer-portal.stage.mdn.mozit.cloud/) - please check the filtering of dates (and topics) works as expected


stevejalim:
@valgrimm (and cc @Gregoor for context )

Thanks @valgrimm for your point over email about the ordering of the events shown by default. 

I've just updated the code so that we show the "next to happen" as the first thing in the list:

Screenshot 2020-06-10 at 10 18 28

This makes more sense for the default "what is happening in the next 6 months?" view, but it does come with a slight UX pain point in that I have also chosen to keep that chronological ordering consistent across all ways to view the Events list by date. This means that currently users may be a bit surprised (or maybe not, depending on one's expectations):

* Default view (next 6 months): shows the soonest-to-happen at the top of the list (a good thing, I think)
* All Future Events: shows the soonest-to-happen at the top of the list (also good thing)
* All Past Events: shows the first-to-have-occurred at the top of the list (less of good thing? or fine?)
* Both Future and Past events selected (ie, shows all events): shows the first-to-have-occurred at the top of the list (less of good thing? or fine?)

What do you think? It might be fine, in real terms (and we can user-test it anyway), but the only other way I can think we handle this (without changing the UI to include an explicit "order-by" option...) would be to set the ordering according to the best fit for the data, even if it is then inconsistent across filters:

No change from _currently_ staged behaviour (ie, chronologically ordered)

* Default view (next 6 months): shows the soonest-to-happen at the top of the list
* All Future Events: shows the soonest-to-happen at the top of the list

A change from the currently staged behaviour, switching to reverse-chronological:
* All Past Events: shows the youngest/most-recently-to-have-occurred at the top of the list (ie, not consistent with the above)
* Both Future and Past events selected (ie, shows all events): show the farthest-in-the future at the top of the list

Implementing the above is easy, too, so it's just a question of which UX do we prefer.

(Adding an explicit order-by filter is not off the table - but it does make the filtering options busier -- happy to discuss that, too)

valgrimm:
Hi, so I'll repeat what I understand and make sure I've got this right.

Steve on event ordering options:
> No change from currently staged behaviour (ie, chronologically ordered)
> - Default view (next 6 months): shows the soonest-to-happen at the top of the list
> - All Future Events: shows the soonest-to-happen at the top of the list

This is my preferred option, if the ordering behavior should be consistent across filter views. 

Summary of Gregor on presentation of default event selection:

- Tick the "Events in the next 6 months" box by default so folks arriving on the page can see that it is prefiltered.

Yes, I agree.

Do you have the information you need?

stevejalim:
@valgrimm Thanks for the approval. I'll leave the ordering as-it-is on staging at the mo (regular chronological). I'll see how quickly I can drop in Gregor's checkbox suggestion, because I like it too

stevejalim:
Splitting out Gregor's nice suggestion to a separate ticket: #1561

Issue #1514, opened 2020-05-28
Update automatic importer to ensure there are descriptions on the items imported, not just titles


stevejalim:
This is so we get more relevant search results for `ExternalArticle`s and `ExternalPost`s

AC
- `ExternalPost` pages that result from the importer have their description set to an appropriate field from the import, such as the RSS summary

- `Video` pages that result from the importer have their description set to an appropriate field from the import, such as the RSS summary IF it is short (<200 chars), else that content is used to set the `Page.body`

PR #1570, opened 2020-06-15
1514: Ingest descriptions from feeds where possible

stevejalim:
This changeset increases the amount of information we use from RSS and Atom feeds when automatically creating `ExternalArticle` and `Video` draft pages, respectively.

In both formats, It looks for a `description` element in the output of the feed parser and, if it contains a `<p>` node, uses that as the `description` field for the page that is auto-created.

Note:
* If there is more than one `<p>` node in the parsed input's `description`, it uses only the first  one
* If there is no `<p>` node, the `description` is not used at all
* If there is a `<p>` node but its contents are longer than 393 characters (ie, the 400 max limit, minus seven characters for `<p>` and `</p>` tags, the content is truncated inside the `p` node is truncated, **so descriptions MUST be read before the page is published**

Examples from local build:

Screenshot 2020-06-15 at 13 29 31

Screenshot 2020-06-15 at 13 29 15

Screenshot 2020-06-15 at 13 33 21
Screenshot 2020-06-15 at 13 33 33

(Related issue #1514)

## How to test

- Code is deployed to Staging, but setting up the appropriate state to import items again without spamming moderators is awkward. So, the next time we get new content to import the importer will set descriptions that can be checked - @valgrimm watch out for a notification email from stage 😄

### DIFF
--- a/developerportal/apps/ingestion/utils.py
+++ b/developerportal/apps/ingestion/utils.py
@@ -19,7 +19,9 @@ from wagtail.core.models import Page
 from wagtail.embeds.blocks import EmbedValue
 
 import feedparser
+from bs4 import BeautifulSoup
 
+from ..common.constants import DESCRIPTION_MAX_LENGTH
 from ..externalcontent import models as externalcontent_models
 from ..mozimages.models import MozImage
 from ..videos import models as video_models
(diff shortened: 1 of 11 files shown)


stevejalim:
> This is great! Would be curious if there is a way to refresh old ingestions with descriptions, but that can come after.

I could certainly script something to do that, as a one-off thing. Question would be how much value it adds given we're at least already indexing the title (because that was present)

valgrimm:
> I could certainly script something to do that, as a one-off thing. Question would be how much value 
> it adds given we're at least already indexing the title (because that was present)

Well, it seems to make sense to be consistent. But that script would definitely be a separate ticket, and a lower-priority one.

stevejalim:
@valgrimm This is running OK on stage now and I can see the most recent article to be auto-imported [here](https://developer-portal.stage.mdn.mozit.cloud/admin/pages/510/edit/) has the description set on it automatically.

developerportal/apps/ingestion/utils.py
```diff
@@ -50,6 +52,41 @@ def _get_item_image(entry) -> str:
...
+
+        # If it's a multi-line description, we only want the first
+        desc = entry.description.split("\n")[0]
+
+        # Extract the string within the first <p>, if any, in the content
+        # (eg this is how Wordpress appears to format it for Mozilla Hacks)
+        soup = BeautifulSoup(desc, features="lxml")
+        try:
+            desc = soup.find("p").text
+        except AttributeError as ae:
```
Gregoor, line 71: I'm of the "don't-use-error-handling-for-that-which-can-be-sanely-checked" school, so I'd personally go with something like:
```python
if text in paragraph:
  desc = paragraph.text
else:
  desc = None
  logger.warning("yadayada")
```

But yours might be more pythonic, I don't know much about pythonicisim

stevejalim: Yep I hear you, and thanks! I went that way because it would be pretty unusual not to get that <p> node, but I realise I do prefer sit on the exceptions-as-control-flow side of the aisle in the Python church, partly because there's (a tiny bit) less code. I think I'll stick with it for now, but I do get your point.

PR #1572, opened 2020-06-15
1546: Trigger Wagtail's search index rebuild on deployment

stevejalim:
This changeset adds k8s deployment config to rebuild search index after deployment.

This is so that if the changeset being deployed contains changes to the search index (including deploying search to production the first time) we need to make sure the index is ready.

It needs to follow after migrations have been completed, because the `wagtailsearch` app itself has migrations.
- summarise changes here, linking to commits as appropriate

## How to test

- This isn't staged yet - for now code review is enough, then after that I will try deploying it to staging. It won't hit production until the related Epic is ready for launch in a couple of weeks' time, but I will ship it to dev and staging before then

### DIFF
--- a/k8s/Makefile
+++ b/k8s/Makefile
@@ -159,6 +159,15 @@ k8s-jenkins-create-rbac:
 k8s-jenkins-delete-rbac:
 	kubectl delete -f jenkins-rbac.yaml
 
+k8s-search-index-update-job: k8s-delete-search-index-update-job
+	env NEW_RELIC_APP_NAME=dev-portal-web-${TARGET_ENVIRONMENT} \
+	j2 search-index-update-job.yaml.j2 | ${KC} apply -f -
+	env JOB_NAME=search-index-update ./wait_for_job.sh
+
+k8s-delete-search-index-update-job:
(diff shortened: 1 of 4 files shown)


k8s/Makefile
```diff
@@ -159,6 +159,15 @@ k8s-jenkins-create-rbac:
 k8s-jenkins-delete-rbac:
 	kubectl delete -f jenkins-rbac.yaml
 
+k8s-search-index-update-job: k8s-delete-search-index-update-job
+	env NEW_RELIC_APP_NAME=dev-portal-web-${TARGET_ENVIRONMENT} \
+	j2 search-index-update-job.yaml.j2 | ${KC} apply -f -
+	env JOB_NAME=search-index-update ./wait_for_job.sh
```
stevejalim, line 165: I set a value for NEW_RELIC_APP_NAME here with the assumption that we'll get output from the command then, rather than it going to the equivalent of `/dev/null` -- is that right @limed ?

limed: Setting the name here just sets  the name of the app (in this case `dev-portal-web-${stage,prod}`) under APM in NR.

Issue #1533, opened 2020-06-03
Event page search: add support for searching as part of refining options


valgrimm:

valgrimm:
Same AC as #1513, but for Events

PR #1575, opened 2020-06-16
1533: Backend support for Events search

stevejalim:
This changeset adds backend support for searching Events.

It uses the same pattern as with Posts page:

    /events/?search=my+query+here

Can also be combined with other parameters

    /events/?topic=css&search=some+keywords

(Related issue #1533)

## How to test

- Code is on [Staging](https://developer-portal-cdn.stage.mdn.mozit.cloud/events/), and you can test it by manipulating the URL, such as above


valgrimm (approved):
Works for me

Gregoor (approved):
🚀

Issue #1581, opened 2020-06-17
Filter form: add support for auto-scrolling to top of results


stevejalim:
At the moment, when you refine results in any page that features the filter-form you get taken to the top of the page (because it's a full-page reload, rather than partial content replacement - for now).

As a quick boost to usability, we should scroll the user to the same point as the pagination controls do (which is a fixed anchor in the page)

PR #1584, opened 2020-06-17
1581: improve the UX of refining a view of Posts, Events or People 

stevejalim:
This changeset adds a change that takes the user to the top of the results list when they "Refine Results" on any page that features the filter form.

(Pagination already did this, so it made sense to also have Refine Results land users on the same part of the page, especially because the default view of an Events page focuses on featured content, not results.

(Related issue #1581)

## How to test

- Code is deployed on [staging](https://developer-portal.stage.mdn.mozit.cloud/)
- Please check filtering and pagination (you may need to deselect filters to get enough content) takes the user to the top of the results:
  - https://developer-portal.stage.mdn.mozit.cloud/posts/
  - https://developer-portal.stage.mdn.mozit.cloud/events/
  - https://developer-portal.stage.mdn.mozit.cloud/communities/people/

I think the experience is better like this, but please say if you disagree

### DIFF
--- a/developerportal/apps/articles/templates/articles.html
+++ b/developerportal/apps/articles/templates/articles.html
@@ -10,8 +10,8 @@
   {% static "img/icons/article-white.svg" as page_icon_asset_url %}
   {% include "molecules/header-strip.html" with content=page.title element="h1" page_icon_asset_url=page_icon_asset_url %}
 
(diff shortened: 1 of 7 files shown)


stevejalim:
This PR also fixes up some wording in comments

Issue #1594, opened 2020-06-19
Events page needs a header above the main events list to aid user navigation


stevejalim:
The Events page differs from Posts and People because it had rich text above the list of cards. As such, particularly on mobile, the page can be confusing to understand.

To fix this we could add a subhead above the lost of even cards, probably only visible at the smallest viewport width (mobile)
```

stevejalim:
@valgrimm one for a chat on Monday

stevejalim:
Handling this as a one-line change in #1539

valgrimm:
Let's discuss before applying any changes.

stevejalim:
Well, it's now rolled in to https://github.com/mdn/developer-portal/pull/1600 because it was a v small change and made the UX of feature delivered in that PR much better, too, so hopefully you'll like it, too. It only appears on the Events page and then only on mobile viewports

PR #1600, opened 2020-06-22
1539: Show applied filters and search params in mobile viewport

stevejalim:
This changeset improves the mobile-viewport UX of filtering and searching the Posts, Events and People pages by showing a summary of any applied filters or search terms above the "Filters" panel, which is collapsed by default in narrow viewports

Screenshot 2020-06-18 at 22 02 48

It also adds in an extra header above the filtering form for the Events page, because the up-page content on Events means that - on mobile - the filter form is a long way away from the page header, so adding in a custom header for Events gives the user a reminder of the context of what's going on. (At least, I think so and hope you agree!) - this issue was reported in #1594

Screenshot 2020-06-22 at 11 15 29

Finally, it also fixes an oversight in configuring Events for search in #1530 -- the `country` field also needed specifying as a wagtailsearch `FilterField` to prevent a 500.

(Related issues #1530 and #1594)

## How to test

- The code has been deployed to staging. 

- Please try using the filters and search parameters on all of the the following URLs in responsive mode (https://developer.mozilla.org/en-US/docs/Tools/Responsive_Design_Mode) where you will see the summary and then in desktop viewport, where you will not see it (nor the extra "Events" header above the filtering)

    - https://developer-portal.stage.mdn.mozit.cloud/events/
    - https://developer-portal.stage.mdn.mozit.cloud/communities/people/
    - https://developer-portal.stage.mdn.mozit.cloud/posts/

- Try adding content that needs escaping (eg `<script>alert("boo");<script>` or an accented name or term) into the search box and confirm that it is shown correctly in the summary of search terms and filters. View source if you can to confirm that it is rendered out as escaped characters (eg `&lt;script&gt;` etc)

## Screenshots

Screenshot 2020-06-22 at 11 10 41
Screenshot 2020-06-22 at 11 14 31
Screenshot 2020-06-22 at 11 15 29


valgrimm:
Tested all in rmd.

Observations all pages:
UX confusion.
Clicking an X doesn't clear the filter, but clicking the word "Clear" does. Perhaps leave out the X and put a border around "Clear".
Clearing filters and search terms does not remove them from the light-gray search info text at the top of the search area, which can make one think that the filter hasn't cleared. 

Search works as expected on all pages. On people of course there are still just filters and that's fine, as people search is out of scope except for full site search.

Did not find anything of note while testing for escaped search or accented letter,  but second set of eyes will be helpful there.

stevejalim:
@valgrimm Thanks for the review

> Clicking an X doesn't clear the filter, but clicking the word "Clear" does. Perhaps leave out the X and put a border around "Clear".

That's interesting... I can't replicate that myself. What browser and OS were you using? Clicking the X works for me. Also, the clear button is not an addition here, so I'd be tempted to split out the fix/amendment as a separate (bug) card, if that's OK with you.

> Clearing filters and search terms does not remove them from the light-gray search info text at the top of the search area, which can make one think that the filter hasn't cleared.

GREAT CATCH. I hadn't appreciated that, and will add it in now. Will ping you for re-review when it's ready.

valgrimm:
> That's interesting... I can't replicate that myself. What browser and OS were you using?

Nightly, Windows 10.

stevejalim:
> > That's interesting... I can't replicate that myself. What browser and OS were you using?
> 
> Nightly, Windows 10.

I can't replicate this via Browserstack, so will have to let it go I think.

stevejalim:
> Clearing filters and search terms does not remove them from the light-gray search info text at the top of the search area, which can make one think that the filter hasn't cleared.

@valgrimm I've updated staging with this behaviour, but to be honest am not sure it's as useful as I thought it may be. If we re-filtered immediately/automatically, it would be essential, but right now it means the summary is updated to match the state of the filtering form, but we actually lose the info that tells the user what the params were that caused the current set of results to be shown.

So, please do take a look, and see if you agree that it might be better to remove this last addition and keep it as it was when you first reviewed it. Cheers!

valgrimm:
I think this is a good question for @malqinneh.
And no, I won't bug you all the time Mustafa (:  but where there is strong uncertainty on significant UX I'd prefer to defer to your expertise. Ideally we won't have a lot more changes.

Basically, the question is: to clear or not to clear?
When a filter and/or search term is removed, should we remove the info text that indicates what was previously filtered/searched for?

stevejalim:
Having slept on it, I still think we should drop it - it's more confusing to change the state of the summary without changing the state of the results: the summary needs to be a summary of what led to the results you see not a summary of what's currently selected in the form.

valgrimm:
Okay.

stevejalim:
Reverting and updating stage

developerportal/apps/topics/tests/test_topics.py
```diff
@@ -44,6 +44,14 @@ class TopicsTests(PatchedWagtailPageTests):
...
     fixtures = ["common.json"]
 
+    def setUp(self):
+        # purge any Topic titles we unintentionally cache
+        cache.clear()
+
+    def tearDown(self):
+        # purge any Topic titles we unintentionally cache
+        cache.clear()
+
```
stevejalim, line 54: This is a fix to a test regression that shows up when re-running tests that now get Topics

stevejalim:
@valgrimm @Gregoor Giving this one a little bump for [re]approval as needed

stevejalim:
Whoops - I'd overwritten the change on stage so it wasn't there to re-test. My apologies! When you get time, please do take another look.

Re auto-submission on filtering, we could do that, sure - but it might cause user pain because it assumes they only want to set one thing before triggering a refinement (and having the user wait for the results before the filter again). See how frustrating (or not!) you find https://www.sportsshoes.com/products/ when trying to filter by two things at the same time (eg for Tops and T-Shirts, or two different brands of shoe). Happy to discuss! We could make the whole thing slicker, but we've been avoiding a big refactor here for reasons we've discussed in Zoom in the past

src/js/organisms/filter-form.js
```diff
@@ -74,24 +107,31 @@ module.exports = class FilterForm {
...
     e.preventDefault();
-    const { controls } = e.target.dataset;
-    const checkboxes = this.form.querySelectorAll('input[type=checkbox]');
-    const matchedCheckboxes = Array.from(checkboxes).filter(
-      (checkbox) =>
-        (!controls || checkbox.name === controls) && checkbox.checked,
+    const triggerEl = e.target.closest('a.js-filter-clear');
+    const { controls } = triggerEl.dataset;
+    const formInputs = this.form.querySelectorAll(
+      'input[type=checkbox], input[name="search"]',
```
schalkneethling, line 115: If we have the search input be a `type="search"` then we can change that here and not be concerned about the value of `name` changing from under us ;)

src/js/organisms/filter-form.js
```diff
@@ -74,24 +107,31 @@ module.exports = class FilterForm {
...
-
-    matchedCheckboxes.forEach((checkbox) => {
-      // eslint-disable-next-line no-param-reassign
-      checkbox.checked = false;
+    const matchedFormInputs = Array.from(formInputs).filter((input) => {
+      return (
+        // if there is a control group specced, is it the one we want to target?
+        (!controls || input.name === controls) &&
+        // and is the input called search or is it a checked checkbox?
+        ((input.name === 'search' && input.value) || input.checked)
```
schalkneethling, line 122: Ditto here

src/js/organisms/filter-form.js
```diff
@@ -74,24 +107,31 @@ module.exports = class FilterForm {
...
+      );
     });
 
-    if (matchedCheckboxes.length) {
-      const event = new Event('change');
-      this.form.dispatchEvent(event);
-    }
+    matchedFormInputs.forEach((input) => {
+      if (input.name === 'search') {
```
schalkneethling, line 127: And here

src/js/polyfills.js
```diff
@@ -196,3 +196,22 @@ if (typeof Object.assign !== 'function') {
...
+  Element.prototype.closest = function (s) {
+    var el = this;
+
+    do {
+      if (Element.prototype.matches.call(el, s)) return el;
+      el = el.parentElement || el.parentNode;
+    } while (el !== null && el.nodeType === 1);
+    return null;
+  };
+}
```
schalkneethling, line 217: Are you use Babel to transpile the JS? If so, and you have included IE11 in the support matric it might already be doing this?

src/js/tests/filter-form-mock-html.js
```diff
@@ -0,0 +1,176 @@
...
+
+        <header class="filter-form-section js-filter-form-clear-section">
+          <a href="#" class="filter-form-clear js-filter-clear">
+            <span class="icon">
+              <svg width="12" height="12" xmlns="http://www.w3.org/2000/svg"><path d="M6 4.586L10.243.343a1 1 0 1 1 1.414 1.414L7.414 6l4.243 4.243a1 1 0 1 1-1.414 1.414L6 7.414l-4.243 4.243a1 1 0 1 1-1.414-1.414L4.586 6 .343 1.757A1 1 0 1 1 1.757.343L6 4.586z" fill="currentColor" fill-rule="evenodd"></path></svg>
+
+            </span>
+            Clear all
+          </a>
+        </header>
```
schalkneethling, line 17: I would probably just use a `button` element here and no use the header at all. A `header` element without any `H` elements seems a little non-semantic ;)

src/js/tests/filter-form-mock-html.js
```diff
@@ -0,0 +1,176 @@
...
+          <header class="filter-form-section-header">
+            <h5>Products &amp; Technologies</h5>
+            <a href="#" class="filter-form-clear js-filter-clear" data-controls="topic">
+              <span class="icon">
+                <svg width="12" height="12" xmlns="http://www.w3.org/2000/svg"><path d="M6 4.586L10.243.343a1 1 0 1 1 1.414 1.414L7.414 6l4.243 4.243a1 1 0 1 1-1.414 1.414L6 7.414l-4.243 4.243a1 1 0 1 1-1.414-1.414L4.586 6 .343 1.757A1 1 0 1 1 1.757.343L6 4.586z" fill="currentColor" fill-rule="evenodd"></path></svg>
+
+              </span>
+              Clear
+            </a>
+          </header>
```
schalkneethling, line 29: Same here as above

src/js/tests/filter-form-mock-html.js
```diff
@@ -0,0 +1,176 @@
...
+            </label>
+
+            <label>
+              <input type="checkbox" name="topic" value="javascript">
+              JavaScript
+            </label>
+
+            <label>
+              <input type="checkbox" name="topic" value="unused-test-topic">
+              Test topic with no logo
```
schalkneethling, line 58: Hmmmm, is this one meant to be in here?

src/js/tests/filter-form-mock-html.js
```diff
@@ -0,0 +1,176 @@
...
+        <fieldset class="filter-form-section">
+          <header class="filter-form-section-header">
+            <h5>Search</h5>
+              <a href="#" class="filter-form-clear js-filter-clear" data-controls="search">
+              <span class="icon">
+                <svg width="12" height="12" xmlns="http://www.w3.org/2000/svg"><path d="M6 4.586L10.243.343a1 1 0 1 1 1.414 1.414L7.414 6l4.243 4.243a1 1 0 1 1-1.414 1.414L6 7.414l-4.243 4.243a1 1 0 1 1-1.414-1.414L4.586 6 .343 1.757A1 1 0 1 1 1.757.343L6 4.586z" fill="currentColor" fill-rule="evenodd"></path></svg>
+              </span>
+              Clear
+            </a>
+          </header>
```
schalkneethling, line 82: Same comment as earlier

src/js/tests/filter-form-mock-html.js
```diff
@@ -0,0 +1,176 @@
...
+          <input class="filter-form-search js-search-input" name="search" value="" placeholder="Refine with keywords">
```
schalkneethling, line 83: This could be a search type input
```suggestion
          <input type="search" class="filter-form-search js-search-input" name="search" value="" placeholder="Refine with keywords">
```

src/js/tests/filter-form-mock-html.js
```diff
@@ -0,0 +1,176 @@
...
+    <div class="filter-list-sidebar-content filter-list-sidebar-content-desktop">
+      <form class="filter-form js-filter-form" data-controls="article_or_video-cards" action="/posts/">
+        <header class="filter-form-section js-filter-form-clear-section">
+          <a href="#" class="filter-form-clear js-filter-clear">
+            <span class="icon">
+              <svg width="12" height="12" xmlns="http://www.w3.org/2000/svg"><path d="M6 4.586L10.243.343a1 1 0 1 1 1.414 1.414L7.414 6l4.243 4.243a1 1 0 1 1-1.414 1.414L6 7.414l-4.243 4.243a1 1 0 1 1-1.414-1.414L4.586 6 .343 1.757A1 1 0 1 1 1.757.343L6 4.586z" fill="currentColor" fill-rule="evenodd"></path></svg>
+            </span>
+            Clear all
+          </a>
+        </header>
```
schalkneethling, line 102: Also here.

src/js/tests/filter-form-mock-html.js
```diff
@@ -0,0 +1,176 @@
...
+          <header class="filter-form-section-header">
+            <h5>Products &amp; Technologies</h5>
+            <a href="#" class="filter-form-clear js-filter-clear" data-controls="topic">
+              <span class="icon">
+                <svg width="12" height="12" xmlns="http://www.w3.org/2000/svg"><path d="M6 4.586L10.243.343a1 1 0 1 1 1.414 1.414L7.414 6l4.243 4.243a1 1 0 1 1-1.414 1.414L6 7.414l-4.243 4.243a1 1 0 1 1-1.414-1.414L4.586 6 .343 1.757A1 1 0 1 1 1.757.343L6 4.586z" fill="currentColor" fill-rule="evenodd"></path></svg>
+
+              </span>
+              Clear
+            </a>
+          </header>
```
schalkneethling, line 114: And here :) Unfortunate that we have this duplication 😢

src/js/utils.js
```diff
@@ -24,20 +23,14 @@ exports.parseQueryParams = () => {
...
-  });
-  return filter;
+exports.decodeFormURLEncodedSpaces = (value) => {
+  /* Turn "foo+bar" into "foo bar".
+   * Context: decodeURIComponent (used above in parseQueryParams) does not
+   * convert `+` to spaces, because the `+` is from the spec for
+   * application/x-www-form-urlenconded, not encodeURIComponent)
+   *
+   * @param {string} value
+   * @returns {string}
```
schalkneethling, line 33: Could be useful to add an example of the value coming in and then how it would be returned

developerportal/templates/molecules/filter-form.html
```diff
@@ -104,6 +104,24 @@ <h5>Country</h5>
...
+  {% if filters.show_search_input %}
+    <fieldset class="filter-form-section">
+      <header class="filter-form-section-header">
+        <h5>Search</h5>
+        <a href="#" class="filter-form-clear js-filter-clear" data-controls="{{SEARCH_QUERYSTRING_KEY}}">
+          <span class="icon">
+            {% include "atoms/icons/close.svg" %}
+          </span>
+          Clear
+        </a>
```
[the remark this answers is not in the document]
stevejalim: Yep - agreed. Ticketed in backlog, but I should address it now. (Pre-Steve code)

developerportal/templates/molecules/filter-form.html
```diff
@@ -104,6 +104,24 @@ <h5>Country</h5>
...
+      </header>
+      <label for="id_search" class="visually-hidden">
+        Search
+      </label>
+      <input id="id_search" class="filter-form-search js-search-input" name="search" value="" placeholder="Refine with keywords"/>
```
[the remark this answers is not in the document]
stevejalim: I think this was added on the base branch, but this PR wasn't rebased on it. Either way, will sort out.

src/css/organisms/filter-form.scss
```diff
@@ -56,6 +56,28 @@
       padding-bottom: 0px;
     }
   }
+
+  /* - Internet Explorer 10–11
+   - Internet Explorer Mobile 10-11 */
+  :-ms-input-placeholder {
+    font-size: 0.8rem;
+    padding: 4px;
+  }
```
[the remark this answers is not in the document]
stevejalim: No rush, but could you clarify this for me, please? Are you saying there's no need to do this and we just let the legacy browser use its default style sheet? (I'm fine with that, but assumed it was worth making slightly nicer)

src/css/organisms/filter-form.scss
```diff
@@ -56,6 +56,28 @@
...
+  .filter-form-search {
+    padding: $spacing-xs;
+
+    @media #{$mq-md} {
+      width: 100%;
+    }
+    width: 97%;
+    @supports (-webkit-touch-callout: none) {
+      // fix for mobile safari only
+      width: 92%;
```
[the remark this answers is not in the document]
stevejalim: Yeah... that was a fun one work out.

However, there is a full overhaul of the buttons due to land when we merge Mozilla Protocol V12 (which is PRed, but am leaving it till after hard launch), so this might go away then.

src/css/templates/events.scss
```diff
@@ -4,3 +4,11 @@ body.events {
     margin-top: -24px;
   }
 }
+
+.events-filter-header {
+  // we only want to show this on non-mobile viewports
+  @media #{$mq-md} {
+    display: none;
+    visibility: hidden;
```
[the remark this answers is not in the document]
stevejalim: Roger that. It's an old habit I picked up years ago. Question was it ever actually necessary to do both?

src/js/tests/filter-form-mock-html.js
stevejalim: - [ ] @stevejalim `a` -> `button`

src/js/tests/filter-form-mock-html.js
stevejalim: - [ ] @stevejalim `a` -> `button`

Yeah, would be good to remove that duplication (which also makes unit tests for the JS a pain) and use `display: flex` then`flex: column-reverse` where appropriate, to put the filter panel above the content. 

(Or it might just be possible with different CSS classes already in Protocol, thinking about it...) However, I'd prefer to split the de-duplication out to a separate ticket

- [ ] @stevejalim split into ticket

src/css/organisms/filter-form.scss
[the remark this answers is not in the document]
schalkneethling: A general thought/comment. The thing is, once you start with one thing it quickly snowballs into many things and then the CSS is bloated for no real good reason 😸  ⛄

schalkneethling:
@stevejalim I see there are a number of conflicts. Let me know when all of this is on stage and I will give it a whirl. Thank you!

stevejalim:
Closing this PR because if/when it gets resurrected, a fresh start is probably better, given other desired refactors, such as removing the duplicated filter form in the HTML etc

PR #1624, opened 2020-06-26
1471: Site search page MVP

stevejalim:
This changeset adds a first-pass/MVP solution to site search for the Developer Portal.

It adds a standalone search results page which features a search input for refinement of searches, but please not that the general staring point for search will be - as with MDN - a search input in the top-of-page navigation (ticketed as #1454)

The search backend in use is the same Postgres-based one that powers search on [/posts/](https://developer-portal.stage.mdn.mozit.cloud/posts/) and [/events/](https://developer-portal.stage.mdn.mozit.cloud/events/) but in this use case it will be searching all configured fields, not just `title` and `description`. In real terms, that means that the `body` field of `Video` and `Article` are also being searched.

Also not that certain pages are deliberately excluded from the scope of search, because they are pages which hold no/little additional information themselves and are more about linking to pages with 'real' content. These are the Homepage, Posts page, Events listing page, Videos page (hidden/unpublished), Topics page (present but not linked to at /topics/) and People  -- however all the pages they link to (and any pages that are children of those linked pages, etc) should appear in search

Results are rendered as links without image previews, but do include the domain of the site that the result will take the user to, because the portal features both internal and external content in its Posts, Events and other pages.

This was all created without a formal design, and all UX and usability feedback is particularly welcome.

Here's a quick couple of screenshots. More follow below:

Screenshot 2020-06-26 at 14 01 50

Screenshot 2020-06-26 at 13 40 38

**Please note: now that PR https://github.com/mdn/developer-portal/pull/1629 has added a search input in the nav, the search input on the results page (ie, this page) is be shown in mobile viewport, and will be hidden in deskport viewport. This is taking a cue from MDN, which has a similar experience.** 

In addition, there are some points that I'd appreciate you bearing in mind as you review and test drive (ideally in your pickiest, most exacting mood 😄 )
 
- Do we need a small amount of copy below the h1 to clarify what is being searched here? the portal both hosts content and also links out to other Mozilla (and third-party) entities
- Should we do more to highlight external links or is the domain enough? Is showing the domain confusing?
- Is it worth adding a role string to the label which says "Person" eg  "Person - Tech Speaker" or even just "Tech Speaker" (cleaner, better) 
- On the Posts page, we feature both Article and Video content in the list of Posts. In the context of search, though, the label "Post" might be too generic and the greater specificity of "Article" or "Video" might be better here.
- We play externally-hosted videos in lightbox, just like we do in other pages on the site. However, if the user back-arrows instead of closing the modal, it navigates back a page. Is this a big deal to you? (I'm wondering about user expectation vs "technically correct")
- The in-page search box has no (visible) search submission button, nor a clear trigger. I took the cue from MDN for these, but would welcome your views on whether they are necessary.
- There's a minimal amount of cacheing in play right now, partly because the CDN should be taking the load. However, if you see any code which you think would benefit from redis-backed cacheing right now, please do say!
- There are also some code-specific comments which I'll add as annotations to the diff.

## How to test

- Code is up on [staging](https://developer-portal.stage.mdn.mozit.cloud/search/) - you can test queries via the input on the page

- [ ] Find multiple matches, including confirming pagination is working
- [ ] Find single results (eg by searching for a very specific string from another page)
- [ ] Find no results (eg https://developer-portal.stage.mdn.mozit.cloud/search/?search=dkjhdflgkjhdglkds)
- [ ] Confirm JS code won't be executed if submitted - eg search for `<script>alert("uh oh");</script>`
- [ ] Try responsive mode - note how the search input flow above the results -- is that better or worse than having it at the bottom?
- [ ] Pick some content that's deeply nested in the site and confirm you can find it via Search

## Screenshots

Screenshot 2020-06-26 at 13 56 41
Screenshot 2020-06-26 at 13 56 49
Screenshot 2020-06-26 at 14 01 50
Screenshot 2020-06-26 at 14 02 05
Screenshot 2020-06-26 at 14 03 37


src/img/icons/search-white.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
stevejalim, line 2: - [ ] @stevejalim to remove doctype

src/img/icons/search-black.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
stevejalim, line 2: - [ ] @stevejalim to remove doctype

developerportal/apps/search/templates/site-search.html
```diff
@@ -0,0 +1,37 @@
...
+      mzp-l-sidebar-right
+      {% if not search_results %}no-results{% endif %}
+    ">
+
+    <main role="main" class="mzp-l-main custom-width">
+      {% include "organisms/search-results.html" with total_results=total_results search_query=search_query search_results=search_results %}
+    </main>
+
+    <aside class="mzp-l-sidebar custom-width">
+      <form class="site-search-form" action="{% url 'search.site_search' %}" method="get">
```
stevejalim, line 27: I did this with no CSRF token here because it's a GET with no side effects – _however_ as I write this comment I'm remembering we to note search terms as "hits" for the search backend, so someone flooding us with bad search terms might distort things. However, we'd then end up with the CSRF token being passed as a GET querystring param, which is a bit mucky. Any opinions on this @Gregoor ? My preference right now is to go with tightness over prettiness

developerportal/apps/search/utils.py
```diff
@@ -0,0 +1,40 @@
...
+            if instance:
+                page_ids_to_exclude.append(instance.id)
+
+        page_ids_to_exclude.sort()
+
+        cache.set(
+            PAGE_IDS_TO_EXCLUDE_CACHE_KEY,
+            page_ids_to_exclude,
+            settings.CACHE_TIME_MEDIUM,
+        )
```
stevejalim, line 38: Given this list won't change very much (if at all, because the pages involved are all key structural ones) it seemed a decent candidate for caching

stevejalim:
@Gregoor Please put on your most sceptical performance-bottleneck hat and we can work out if there are any tricky bits to sort out

stevejalim:
@valgrimm Do you think we should index the sidebar on ContentPages too? If left alone I would, but checking in case you think it'll add more noise than usefulness

valgrimm:
I agree with you

developerportal/apps/search/templates/site-search.html
stevejalim: Also, this is relevant: https://infosec.mozilla.org/guidelines/web_security#csrf-prevention

Gregoor: I think for cross-site worries CSRF is more of a fallback at this point anyway, given that SameSite cookie support is at [~90%](https://caniuse.com/#feat=same-site-cookie-attribute) (and probably at 99.99% for developers visiting the site). So imo GET without CSRF is fine.

Then again, I don't know much about how that hit counting is working. Can you link me to some code I could use to read-up on it?

developerportal/apps/search/views.py
```diff
@@ -0,0 +1,46 @@
...
+from ..common.constants import PAGINATION_QUERYSTRING_KEY, SEARCH_QUERYSTRING_KEY
+from ..common.utils import paginate_resources, prep_search_terms
+from ..search.utils import get_page_ids_to_omit_from_site_search
+
+SEARCH_RESULTS_PAGE_SIZE = 15
+
+
+def site_search(request, template_name="site-search.html"):
+    "Whole-site search view"
+    search_query = request.GET.get(SEARCH_QUERYSTRING_KEY, None)
```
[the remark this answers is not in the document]
Gregoor: I think it's fine too! Imo those Form classes are useful for putting validations into a single place and reusing them. The latter doesn't seem feasible here and given that it comes at the cost of more code and indirection, my "calculcation" comes to the same result as yours 🙃

data/devportal_topic_icons/search.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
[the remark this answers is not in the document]
schalkneethling: I would be sure to run them all through SVGOMG

developerportal/apps/search/templates/site-search.html
```diff
@@ -0,0 +1,37 @@
...
+        <label for="id_query">
+          Search this site
+        </label>
+        <input class="site-search-field" id="id_query" type="text" name="{{SEARCH_QUERYSTRING_KEY}}" placeholder="Enter search terms" value="{{ search_query|default:'' }}">
+        <button class="visually-hidden" type="submit">Search</button>
```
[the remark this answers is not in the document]
schalkneethling: MDN is incorrect 😱 I would keep it. I would also suggest setting the type of the input to `search`

developerportal/templates/molecules/search-result.html
```diff
@@ -0,0 +1,39 @@
...
+    {% endif %}
+  >
+  <h4>
+      <span class="search-result-title">{{ page.title }}</span>
+      {% with page|page_type_label as label %}
+        {% if label %}
+          <span class="search-result-type">{{label}}</span>
+        {% endif %}
+      {% endwith %}
+    </h4>
```
schalkneethling, line 24: Do we need both `span` elements? Could the main styling be on the `h4` and then the nested `span` overrides as needed?

developerportal/templates/organisms/search-results.html
```diff
@@ -0,0 +1,30 @@
+<div id="results-list" class="search-results">
+{% if search_results %}
+  <h4 class="search-results-summary">
+    {{total_results}} result{{total_results|pluralize}} found for "{{search_query}}"
+  </h4>
+  {% if total_results > 1 %}
+  <p class="pagination-summary">
+    Showing results {{search_results.start_index}} to {{search_results.end_index}}
+  </p>
+  {% endif %}
```
schalkneethling, line 10: Without having interacted with the actual page/functionality, it feels like this can be an `article` element and the following two can wrapped as an `hgroup`. Something like:

```
<article id="results-list" class="search-results" aria-labelledby="results-heading">
{% if search_results %}
  <hgroup id="results-heading>
    <h4 class="search-results-summary">
      {{total_results}} result{{total_results|pluralize}} found for "{{search_query}}"
    </h4>
    {% if total_results > 1 %}
      <h5 class="pagination-summary">
        Showing results {{search_results.start_index}} to {{search_results.end_index}}
      </h5>
    {% endif %}
  </hgroup>
```

That way we create an explicit section and the heading exposed to assistive technologies would be something like:

```
10 results found for "iot" Showing results 1 to 5
```

if there is only 1 result it would be identified as:

```
1 result found for "iot"
```

This does also mean that `aria-labelledby="results-heading"` needs to wrapped in a conditional to ensure we only set the `hgroup` as the label for the section when there are actual results.

developerportal/apps/search/utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+    page_ids_to_exclude = cache.get(PAGE_IDS_TO_EXCLUDE_CACHE_KEY, [])
+
+    if not page_ids_to_exclude:
+
+        for model in (Articles, Events, HomePage, People, Topics, Videos):
+
+            # There will be only one of each of these pages, but we don't want to
+            # use get() as that would blow up site search if one section was removed
+            instance = model.objects.first()
```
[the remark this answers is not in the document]
stevejalim: Ah it's not the first instance of a page we want to show, it's actually the singleton parent page of a type of pages we want to get and ignore in results

eg we have a parent `Articles` which has many `Article` pages (note the singular) as children. (I didn't name these models!), ditto for `Events > Event`, `Topics > Topic` and `Videos > Video`

developerportal/templates/molecules/search-result.html
stevejalim: True! Will revisit!

developerportal/apps/search/utils.py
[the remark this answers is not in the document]
Gregoor: Aah right, forgot about that pattern here. A little confusing but I guess it's the cards we've been dealt. Thanks!

developerportal/apps/search/templates/site-search.html
stevejalim: https://github.com/wagtail/wagtail/blob/bb2e460c0b92fba802fc3f369730004d6b648e64/wagtail/search/models.py#L20 is where Query.add_hit() is defined. Looks like it "just" logs to a table/model, `QueryDailyHits`, but it doesn't look like much is done with that by default... https://github.com/wagtail/wagtail/search?q=QueryDailyHits&unscoped_q=QueryDailyHits

I'll check in with the Wagtail team and see what they say about this

developerportal/templates/molecules/search-result.html
stevejalim: Ah ha - I remembered/rediscovered why i needed the two spans -- it was to stop the reflow breaking down on mobile viewport:

Without the span for the title text, I have to put `left-margin` on the `span.label` instead to space it away from the H4 main text. For example, in med and large viewports:

Screenshot 2020-06-30 at 22 40 00

but this breaks down particularly on the smallest viewport if the margin is on the `span.label` rather than on a span wrapping the main h4's text:

No  `span.search-result-title`:

Screenshot 2020-06-30 at 22 35 05


With `span.search-result-title`:
Screenshot 2020-06-30 at 22 35 31

developerportal/apps/search/templates/site-search.html
stevejalim: So it looks like it populates data for Wagtail's promoted search behaviour,  which we're not using in devportal. BUT it also tunes the results of some of the searches in the CMS, so it might be that an Editor could be shown less relevant results if there's a dedicated poisoning campaign, but I'm not sure how much of a real pain that'd be...

developerportal/templates/organisms/search-results.html
stevejalim: Thanks! Change made and I'll stage it tomorrow AM

developerportal/templates/molecules/search-result.html
```diff
@@ -0,0 +1,39 @@
...
+    {% if page.resource_type == "video" and page.is_external %}
+      js-modal-trigger{% endif %}"
+    {% if page.is_external %}
+      target="_blank"
+      rel="nofollow noopener"
+      data-title="{{ page.title }}"
+      data-type="{{ page.resource_type }}"
+    {% endif %}
+  >
+  <h4>
```
schalkneethling, line 17: I notice that we jump from an `h1` element to an `h4`. Not a super big deal I guess, it would be better if these were `h2` elements though. The styling can remain the same in terms of font size

schalkneethling:
Perhaps it is just me, but it feels a little odd to have the pagination offset to the right on desktop:

Screenshot 2020-07-01 at 12 44 26

Feels more natural when aligned to the left with the rest of the content:

Screenshot 2020-07-01 at 12 44 26(b)

It looks like the change is to remove `text-align: center;` from `.pagination-block`

developerportal/templates/molecules/search-result.html
schalkneethling: Could we have:

```
span.search-result-title {
    margin-left: 0;

    mq-tablet-and-up {
	    margin-left: 10px;
    }
}
```

schalkneethling:
The color contrast of the little labels, "Post", "Event", etc. fails WCAG. Do we have a slightly darker grey? Something close to `#565662` that we can use as the background color?

developerportal/apps/search/templates/site-search.html
stevejalim: I think, given we're not using promoted search (and have no imminent need to), there's no need to log search hits, so I'll just drop that line of code

- [ ] @stevejalim to drop hit counting

developerportal/templates/molecules/search-result.html
stevejalim: @schalkneethling Tried it but still not quite idea, because smallest viewport ends up with not enough left-margin for the labels
Screenshot 2020-07-01 at 23 14 11
 

Am thinking I might have to just live with this markup for now, if that's OK. Can ticket it to improve when I have more time

stevejalim:
@schalkneethling 

> The color contrast of the little labels, "Post", "Event", etc. fails WCAG. Do we have a slightly darker grey? Something close to #565662 that we can use as the background color?

Great point! Embarrassed to have overlooked that. Previously it was Marketing Gray 60 from https://protocol.mozilla.org/fundamentals/color.html - I've upped it to Marketing Gray 70, which is slightly darker at `#5e5e72`

developerportal/templates/molecules/search-result.html
stevejalim: Done - cheers!

stevejalim:
OK - all points addressed as far a I can right now. Will re-stage the code shortly. What do you think now @schalkneethling ?

stevejalim:
@valgrimm I can see what @Gregoor is saying about the big Search strip in the results page. I was trying to balance consistency with the rest of the site, but maybe that doesn't work. 

For now, shall we go with the current cut of things, and update it when we've had time to snag @malqinneh for some UX feedback too?

stevejalim:
@valgrimm Actually, looking at what https://developer.mozilla.org/en-US/search?q=test does, it might be a simpler solution to retitle the page "Search Results"! I'll do that now and we can see how it looks/helps things

PR #1628, opened 2020-06-27
1623: Support more featured items in topics

stevejalim:
This changeset addresses a need to be able to show more than five featured items on a `Topic` page. 

It now allows up to seven, including six.

This change only applies to the Topic page, but it would be easy enough to use on the Homepage and Events page, if need be, simply by increasing the maximum allowed in the editor (a two-character code change).

The previous layout patterns are still supported, with the same results as prior to the refactor

#### Example of six featured cards

Screenshot_2020-06-27 AV1 Video(1)

#### Example of seven featured cards

Screenshot_2020-06-27 AV1 Video

(Related issue #1623)

## Key changes:

- Refactored Django tempate tag that splits up a list of _n_ pages into (effectively) a list of sub-lists, with one sub-list for each row, including an explicit count of items in each row
- Refactored "featured items" HTML component to support up to four items in a row
- Generally the Python and HTML code are cleaner and less opinonated than before
- Updated config for the `Topic.featured` Streamfield to allow for 7 items, up from 5. Tweaked the help-text in a way that doesn't need an additional migraiton (because changing help-text is a no-op in database terms, it's fine to edit the last migration that captured the state of that help_text)

## How to test

- Code is deployed to DEV. Please use preview mode of a Topic page to easily confirm that 2....7 items are possible and work fine.

### DIFF
--- a/developerportal/apps/common/tests/test_app_tags.py
+++ b/developerportal/apps/common/tests/test_app_tags.py
@@ -264,11 +264,37 @@ class AppTagsTestCase(TestCase):
     def test_split_featured_items(self):
 
         cases = [
-            {"input": [1, 2, 3, 4, 5], "output": ([1, 2], [3, 4, 5], [])},
-            {"input": [1, 2, 3, 4], "output": ([1, 2], [], [3, 4])},
-            {"input": [1, 2, 3], "output": ([], [1, 2, 3], [])},
-            {"input": [1, 2], "output": ([1, 2], [], [])},
-            {"input": [1], "output": ([1], [], [])},
+            {
+                "input": [1, 2, 3, 4, 5, 6, 7],
+                "output": [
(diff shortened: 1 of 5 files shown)


stevejalim:
> I've spent a bit too long trying to figure out whether this should be possible with CSS grid. But I can't seem to come up with a way to do it which makes me a bit sad. @schalkneethling do you, or is this out-of-scope for grid?

It might well be poss, but I'm building off the components in `mozilla/protocol`, so didn't want to pull it all to bits. Would be interesting, though.

stevejalim:
@valgrimm I can deploy this to production tomorrow (with no breaking changes) so it should be ready for you to use as you need

valgrimm:
👍 👍 👍 👍 👍 👍 👍 👍

Issue #1605, opened 2020-06-22
Three featured events on Event page instead of two


valgrimm:
This is to make the featured events cards smaller; in general we will not feature three events

stevejalim:
> This is to make the featured events cards smaller; in general we will not feature three events

I think I'm missing something here: if we allow the layout to take up to three Events in the featured section, they will currently only be rendered smaller if there are three to show, else the layout uses the two-card sizing, ie:

```
[      Event 1     ] [     Event 2    ]
```

Are you needing the featured panel in Events to be like the following?

```
[ Event 1  ]  [ Event 2  ]  [  Blank  ]
```  

If so, we could support the idea of adding a 'Spacer' element to a Featured panel, which would basically be a card with no content and no styling, to fill up the space

stevejalim:
Plan: swap it out for a new block that handles 1...4 items, leaving whitespace rather than auto-resizing

Light gray background

"Featured events" header

If no topic available to provide content for the tag, don't have it - leave it a blank line (whitespace)

PR #1654, opened 2020-07-02
1605 Custom featured events layout

stevejalim:
This changeset addresses the problem that two featured events on the Events page are rendered far too large to be ideal. Instead, we now render up to four Events, but always using a quarter-width card, which are smaller. The overall style of layout is the same as the "What we're working on" panel from Topic pages.

(Related issue #1605)

## Key changes:

- Tweaked the model and last migration to allow more than two featured items for the Events page. It's not a change that affects the database schema, so it's OK to tweak the most recent migration
-  Refactored 'recent work' organism used in Topic page to make it reusable as an organism called 'four-card-row'. By default each item in the row of four will try to show the target page's primary topic as the tag content, else defaults to "Featured". However if show_domain is passed in, the card will get the domain for the outbound link and use that as the tag, preserving the current behaviour for the Topic page. 
- Also made the grey tint background slightly darker.
- Refactored Events page to use the just-created four-card-row organism to display its featured items. This means that there may be empty space if there are fewer than four items to be shown, but this is the requested behaviour. ie, for this page, there is no longer automatic layout adjustments according to the number of elements.
- I've also limited the max width of the rich-text blocks to 768px to improve readability of long lines of text in them.

Checks:
- HTML validated
- Cross-browser checked

## Screenshots both with and without "top page content" above the Featured items

Screenshot_2020-07-02 Events(2)
Screenshot_2020-07-02 Events(3)

Screenshot_2020-07-02 Events(1) 

Screenshot_2020-07-02 Events

## How to test

- Code is on its way to Staging.
- Try configuring Featured Events with and without top content.
- Notice, too, that the width of the top/middle/bottom content rich text is now made narrower, ideally improving readability


src/css/templates/events.scss
```diff
@@ -1,6 +1,24 @@
 body.events {
-  div.events-extra-content {
-    padding-top: 0px;
-    margin-top: -24px;
+  div.header-strip {
+    margin-bottom: 0px; // to allow for the four-card-row div of featured items
```
schalkneethling, line 3: No units with 0 values. @stevejalim does this project use Styellint? Would be useful to set that up, as it will catch these early. Also, I believe Prettier might even fix this automatically. I guess the other question is, does this project use Prettier?

If you want to add those two, you can use the following to easily add the needed config files etc.

https://github.com/project-calavera/project-calavera

stevejalim: Thanks @schalkneethling - the project does use prettier, in its default configuration only, which reformats/indents, etc but doesn't seem to catch this. Good call about stylelint - I've ticketed it as #1666, because it might be a largish diff to fix up any CSS style issues

stevejalim: As for the nested divs, I know what you mean. It likely could be done cleaner, but I'm trying to avoid messing with the partials I've drawn out of protocol.mozilla.org so that breaking upgrades to Protocol are easier to handle. If i get time, I can look to do a "div-reduction" sweep of the codebase, ideally...

Issue #1643, opened 2020-07-01
As a site visitor, I want to know which events are online


valgrimm:
AC: Add an "Online" checkbox to the location filters

stevejalim:
Do you want "Online" to be one of the options in the (currently named) Countries section? (If so, we should rename that Locations, and perhaps we should do that _anyway_)

Or would you like "Online" to be a checkbox in a standalone section? If so, I wonder what we could call that section? We could call that section "Type" and then have "Online" and "In person" as the two options

valgrimm:
> Do you want "Online" to be one of the options in the (currently named) Countries section? (If so, we should rename that >Locations, and perhaps we should do that anyway) ✔️ 

As above. Kudos!

stevejalim:
Ballparking a 1pt, conservatively, as this might be as simple as adding a fake "Online" country to the options, or it might involve going down the rabbit hole

valgrimm:
✔️

PR #1660, opened 2020-07-03
1643: Add "online" as location option

### LINKED ISSUE #1177: Change "Country" fields to Country or Region
Just realized that had not been done, I had read what I thought was there and not what was there. 

per:
https://wiki.mozilla.org/Lists_of_Countries_and_Regions

stevejalim:
This changeset allows us to select "Online" as a location for an Event.

(Please read all the key changes, though)

Related issue #1643. 

Resolves #1177

## Key changes:

- Add support for "Online" as an option an event's "country" as part of the regular drop-down of countries
  - Note that this also makes "Online" an option for a Person too, but we don't have to use it / can just ignore it
- Instead of using the country code as display text in the card for an event and in the header strip for an event, we use the full/verbose name of the country. This is because the country code for the fake "Online" location is `QQ`, which makes no sense as a display string.
- Rephrase 'Country' in the user-facing site to Location (because that encompasses Online, too)
- Update phrasing in the CMS (non-public) to 'Country and Region' as requested in #1177
  - This could be further rephrased to just "Location", maybe?

Screenshot 2020-07-03 at 10 36 09
Screenshot 2020-07-03 at 10 36 22
Screenshot 2020-07-03 at 10 42 31
Screenshot 2020-07-03 at 10 42 37
Screenshot 2020-07-03 at 10 43 52
Screenshot 2020-07-03 at 10 43 57

## How to test

- Code is deployed to Stage
- Confirm that picking a country of "Online" is acceptable UX from an editor's POV
- Add/Edit an Event to have a location of "Online" 
  - confirm the card for the event looks ok / is appropriately labelled
  - confirm the header strip for that Event is also appropriate
  - confirm the location for other events is not broken/impaired
- Confirm filtering, including by location, still works as expected for:
  - Events
  - People


stevejalim:
@valgrimm Please take a look at this when you get time. In particular, I am wondering whether you'll be content with just "Online" as a new country option, and what you think about the rephrasing in the CMS UI.

Once you're happy, I'll call in code review from Gregor

valgrimm:
yes, please change "Country" to "Location" wherever it appears. This supersedes #1177

stevejalim:
@Gregoor Yeah the other uses of country are all checked/swept. THanks!

Issue #1669, opened 2020-07-06
Rate-limiting


stevejalim:
We should rate-limit search queries, partly as a security mitigation and partly to prevent DDOS/saturation that the CDN can't handle for us

AC 
- [ ] rate-limiting is set up
- [ ] the CDN does not stop rate-limiting from working (ie, it looks at the forwarded/end-user IP, not the CDN IP)
- [ ] covers site-wide search
- [ ] covers posts-page search
- [ ] covers event-page search
- [ ] documentation added reminding to add it to new pages that get search functionality

stevejalim:
Kuma uses `django-ratelimit` so we know it's up to the task

PR #1673, opened 2020-07-07
1669: Add rate limiting 

stevejalim:
This changeset addresses the need to rate limit Wagtail pages which support search and the custom site-search view. 

This is needed because pages which process a search input (Posts, Events and Site Search) can accept a huge variety of  search args via GET param, and so these pages will likely **not** already be cached in the CDN (unless someone has searched for the same term since the CDN was last invalidated). Rate limiting will protect us from unnecessary load due to bad actors / misbehaving bots. 

The rate limiting is carried out by adding custom middleware that uses a core (public) helper function from `django-ratelimit`, which is what is also used in Kuma. It rate-limits ALL pages, currently to 25 requests per IP per minute. (Details of the decision are below)

----

While regular Django function- or class-based views can be easily rate-limited via a decorator, it is not possible to do the same in a Wagtail project, because most pages in the site are handled by Wagtail's own routing/resolving logic, which is not easy to decorate. Hence the choice to do it with middleware which can "get in front of the" page serving logic and return early if need be.

This decision also means the _entire_ site is rate-limited - currently at a limit of 25 requests per IP per minute - rather than only the views that the CDN will not be cacheing 

When an IP is rate-limited, we send the user to a custom page

Screenshot 2020-07-06 at 23 52 38

(While in the neighbourhood, I also made a custom 403 page)

Screenshot 2020-07-07 at 00 19 17

---

**Why are we rate-limiting ALL views and not just those that the CDN won't be cacheing? Here's the logic:**

* Most production traffic spikes will be handled by the CDN. The most likely pages
to be CDN cache misses are search ones, because the GET params can contain anything
* Rate-limiting Search is the real goal here, but we have challenges to overcome.
* Search appears on three pages - one is a vanilla Django view `search:site-search`
two are Wagtail pages which are served by `wagtail.core.views.serve`, and those
pages handle GET-based search queries themselves -- they don't hand off to the
`search:site-search` view
* We can't decorate the Wagtail `serve` view with `django-ratelimit`, because it is buried inside Wagtail, but we can use
middleware to do perform rate-limiting checks, using `django-ratelimit`'s core helpers
* However, `django-ratelimit`'s API/helpers need a `HTTPRequest` object, because they
rate-limit based on a view func, _not_ a path from the URL, so we can't easily target
specific Wagtail-served pages for rate limiting.

So: given that non-search pages will be cached by the CDN, we rate-limit ALL the
pages served by the site. In the wild, once the CDN is warm, the majority of
uncached requests will be search ones, so we'll effectively be only rate-limiting
them.

----
(Related issue #1669)

## Key changes:

* Ensure Site Search page is out of scope for spiders - a one-line change that brings portal in line with kuma
* Add django-ratelimter to project + HTTP 429 template, to show to throttled users + add HTTP 403 template while in the neighbourhood
* Add middleware to enforce rate limiting
* Ensure rate-limiting is disabled for health check tests
* Add tests for rate-limiting middleware, to confirm it's using django-ratelimit's internals effectively
* Add in middleware to ensure X-Forwarded-For header is processed appropriately so rate-limiting works on client IP, not CDN or ELB IP. Based on equivalent middleware in Kuma.
* 1669: Add test for the rate_limited HTTP 429 view 

## How to test

This one's less straightforward to test, so I would recommend code review as the main thing, but as far as a check goes, with the knowledge that requests are currently limited to 25/min per IP address on stage (where the code is deployed)

1) Load any page on the non-CDN-fronted staging site repeatedly, 30 times. Your 26th attempt will trigger the rate limiting, for a minute. eg: https://developer-portal.stage.mdn.mozit.cloud/search/?search=test
  * When you get the rate-limited page, use devtools to inspect the response headers and confirm you see the `Cache-Control` set as `max-age=0, no-cache, no-store, must-revalidate` and the HTTP response code as 429

2) Load that same page via the CDN-fronted staging site: https://developer-portal-cdn.stage.mdn.mozit.cloud/search/?search=test and you will find you can hit it again and again, way more than 26 times and not get rate limited, because the CDN is serving the page from its cache. (Responses will be HTTP 200 always and Devtools' network pane will show it as a `Hit from Cloudfront`


stevejalim:
Assigned to @peterbe @escattone and @Gregoor but only needs brainpower from one of you, whoever fancies it!

developerportal/apps/common/middleware.py
```diff
@@ -0,0 +1,87 @@
...
+    latter is set. This is useful if you're sitting behind a reverse proxy
+    """
+
+    def middleware(request):
+        forwarded_for = request.META.get("HTTP_X_FORWARDED_FOR")
+        if forwarded_for:
+            # HTTP_X_FORWARDED_FOR can be a comma-separated list of IPs.
+            # The client's claimed IP will be the first in the list, as CDN etc
+            # append to the end.
+            forwarded_for = forwarded_for.split(",")[0].strip()
```
stevejalim, line 81: This is how Kuma currently does it, but Ryan and I have briefly discussed a more robust approach which I can share in Slack

- [ ] @stevejalim to ticket up a follow-up extension to this middleware

developerportal/apps/common/tests/test_app_tags.py
```diff
@@ -236,6 +236,18 @@ def test_pagination_additional_filter_params(self):
...
+                "input_querystring": (
+                    f"?{TOPIC_QUERYSTRING_KEY}=css&{TOPIC_QUERYSTRING_KEY}=javascript"
+                    f"&{PAGINATION_QUERYSTRING_KEY}=2"
+                    f"&{SEARCH_QUERYSTRING_KEY}=test+string"
+                ),
+                "expected_output": (
+                    f"&{TOPIC_QUERYSTRING_KEY}=css&{TOPIC_QUERYSTRING_KEY}=javascript"
+                    f"&{SEARCH_QUERYSTRING_KEY}=test+string"  # note: still escaped
+                ),
+            },
```
stevejalim, line 250: Stray change from a different branch - will rebase it away

developerportal/apps/events/tests/test_events.py
```diff
@@ -345,6 +345,38 @@ def test_events__get_events__query__all_params_including_future_events(
...
+        expected_q = Q()
+        expected_q.add(countries_q, Q.AND)
+        expected_q.add(overall_date_q, Q.AND)
+        expected_q.add(topics_q, Q.AND)
+
+        events_page.get_events(fake_request)
+        mock_get_combined_events.assert_called_once_with(
+            events_page, q_object=expected_q, search_terms="test here", reverse=False
+        )
+
```
stevejalim, line 379: Again a stray change from another branch - will rebase it away

valgrimm:
broke it once to get the response page, but now cannot break again. Seems fine from my end otherwise as that might be user error. I trust code review folks to uncover any issues.

peterbe:
Seems to work:
```
▶ python dummy.py
1 200
2 200
3 200
4 200
5 200
6 200
7 200
8 200
9 200
10 200
11 200
12 200
13 200
14 200
15 200
16 200
17 200
18 200
19 200
20 200
21 200
22 200
23 200
24 200
25 429
  File "dummy.py", line 6, in <module>
    r.raise_for_status()
  File "/Users/peterbe/.pyenv/versions/3.8.1/Python.framework/Versions/3.8/lib/python3.8/site-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://developer-portal.stage.mdn.mozit.cloud/search/?search=test
```

peterbe:
Screen Shot 2020-07-08 at 8 16 23 AM

developerportal/templates/robots.txt
```diff
@@ -2,3 +2,4 @@ User-agent: *
 Disallow: /admin/
 Disallow: /django-admin/
 Disallow: /static/
+Disallow: /search/
```
peterbe, line 5: I don't think you should do this. I don't see the reason. It's not harmful to allow spiders in there. It's not dangerous.

developerportal/urls.py
```diff
@@ -28,8 +31,17 @@
...
+# Custom 403 handling, sending either a rate-limited response or a regular Forbidden
+def handler403(request, exception=None):
+    if isinstance(exception, Ratelimited):
+        return rate_limited(request, exception)
+    return permission_denied(request, exception)
+
+
 if settings.DEBUG:
     urlpatterns += [
+        url(r"^429/$", rate_limited, {"exception": Ratelimited()}),
```
peterbe, line 43: What are these for? Is it just so you can manually preview, in the browser, what these error pages look like?

developerportal/settings/base.py
```diff
@@ -354,6 +358,14 @@
 # password fields. You'll also need to enable the model backend in local settings
 USE_CONVENTIONAL_AUTH = False
 
+# django-ratelimit - see developerportal.apps.common.middleware.rate_limiter
+RATELIMIT_ENABLE = os.getenv("RATELIMIT_ENABLE", "True") == "True"
+RATELIMIT_USE_CACHE = os.getenv("RATELIMIT_USE_CACHE", "default")
+RATELIMIT_VIEW = "developerportal.apps.core.views.rate_limited"
+DEVPORTAL_RATELIMIT_DEFAULT_LIMIT = os.getenv(
```
peterbe, line 365: Off topic, but you should consider switching to `python-decouple` which is the same as `os.getenv` but it can do automatic casting and it can automatically read for a `.env` file for local development.

developerportal/templates/robots.txt
stevejalim: Noted - I was being paranoid 😄 

- [ ] @stevejalim to remove this

developerportal/urls.py
stevejalim: Yep, exactly! Makes local dev easier and because `DEBUG` is never True in our deployed sites, remains local only

developerportal/settings/base.py
stevejalim: 💯 YES. It's already in the backlog to do and I will be v happy to get to it, if i can

developerportal/apps/common/tests/urls_for_tests.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
```
stevejalim, line 1: Just noticed - we don't need this AND urls defined in the test module. This is a hangover from refactoring

- [ ] @stevejalim to remove this file

developerportal/apps/common/middleware.py
stevejalim: Ticketed as #1686 1686

Issue #1606, opened 2020-06-22
Reduce newsletter footer size on event detail pages


valgrimm:
AC:
- [ ] 40% reduction of image and text size and addition of one footer's worth of white space between the footer and event content, to be more in harmony with page layout

stevejalim:
@valgrimm Could you provide an annotated screengrab of how you want it to look, please?

Also, I get the "Why" here -- and am wondering whether we make this the default size/layout for the newsletter panel on all pages _except for_ the homepage, where it's the current, larger size. What do you think?

Worst-case-ballparking this at 1pt/1 day, but could well be less.

stevejalim:
Plan is:

Make a compact version of newsletter similar to compact call out - https://protocol.mozilla.org/patterns/organisms/call-out.html

We'll use this on all pages apart from homepage, so we need to retain the current version, too

PR #1680, opened 2020-07-08
1606 Newsletter section: compact version

stevejalim:
This adds a more compact rendering of the newsletter form, which is used on the Events page and any Event page

(Related issue #1606)

## Key changes:

-  Refactor the original/main newsletter form ahead of creating compact version.  Still renders the same as before. The JS that the form uses is unaffected by this change
- Support a more compact version of the newsletter organism, which is used only on Event and Events page

## Screenshots

Screenshot 2020-07-08 at 14 15 25
Screenshot 2020-07-08 at 14 15 42
Screenshot 2020-07-08 at 14 54 22

## How to test

- Code will be on staging.
- Take a look at newsletter unit on the Events page and an Event page, ideally across viewport sizes. 
- Also confirm the newsletter unit on the homepage is visually unaffected (eg compare with production)


stevejalim:
@valgrimm @sandrapersing This is up on https://developer-portal.stage.mdn.mozit.cloud/events/ now - let me know what you think

PR #1685, opened 2020-07-09
1621: Add load-testing tooling and perform initial test

stevejalim:
This changeset adds support for using Locust (https://locust.io) as a convenient way to load test the developer portal.

The approach used is a Dockerized solution (from a pre-built container: https://hub.docker.com/r/locustio/locust) which spins up a lead and worker nodes to increase the number of connections a single local machine can make.

If we want to take things further, we could try a Lambda-based Locust runner.

TLDR: I think we're fine for now.

* Dev deployment, with two small pods for `app`

- 100 concurrent users, hatched at 3/sec
  - Without CDN: sustained ~42 requests per second (RPS) and no saturation issues
  - with CDN: reached ~100 RPS and no saturation issues

- 250 concurrent users, hatched at 3/sec
  - included pagination queries (which will hit the origin until they've been cached by the CDN)
    - without CDN, and with rate-limiting turned off, started saturating DB connections ~240 concurrent users
    - with warmed CDN:
    - 100RPS was acheivable once the CDN was warm
    - response times were tiny (~24ms and ~40 for 95th percentile) - thanks, CloudFront!
   - With cold CDN and no rate limiting enabled, I stepped it up 500 simulated users
     - Got up to 215 rps with no noticeable warm-up issues. Initial page load started at 1000ms (which we can improve, but production's greater number of pods might also help with contention), but soon came down to ~100ms 95th percentile, then lower still as time went on.
     - It would be nice to pre-warm the CDN, but this is non-trivial as there are various edge locations to warm

* Staging deployment, with 2x the resources of the dev deployment

  - rate limiting was enabled
  - search was included, as a minority proportion of the requests - see the locustfile.py and the weightings in the @task() decorator
  - 500 users, spawning at 3/sec
    - starting with cold CDN:
       - comfortably hit 215 rps, and triggered rate limits for 1% of requests on search and pagination after ~300 users, but that's because it's from my local IP. 
    - without CDN, Stage got rate limited fast, but didn't 500

### DIFF
--- a/locust/locustfile.py
+++ b/locust/locustfile.py
@@ -1,0 +1,184 @@
+import random
+import string
+
+from locust import between, task
+from locust.contrib.fasthttp import FastHttpUser
+
+
+def get_asset_paths(add_cachebreaker=False) -> list:
+    """Return the JS and CSS bundle paths for the site -- at the moment we have
+    no need for code splitting, etc
+
+    While production uses django-whitenoise to serve files with a cachebreaking
+    hash, the files are also available via their unhashed filename.
+    If we want to, we can ask for them after appending a random cachebreaker
+    param as a querystring to ensure we get a fresh set every time we run the
+    load tests
+    """
+
+    if add_cachebreaker:
+        cachebreaker = f"?v={random.random()}"
+    else:
+        cachebreaker = ""
+
+    return [
+        f"/static/js/bundle.js{cachebreaker}",
+        f"/static/css/bundle.css{cachebreaker}",
+    ]
+
+
+def get_random_string(length=9):
+    return "".join(random.choice(string.ascii_lowercase) for i in range(length))
+
+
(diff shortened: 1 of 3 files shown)


locust/locustfile.py
```diff
@@ -0,0 +1,184 @@
...
+            self.client.get(asset_path)
+
+    @task(2)
+    def view_events_page_filtered(self):
+        random_country_code = random.choice(["GB", "US", "NO", "NL", "DE"])
+        path = f"/events/?date=past&date=future&location={random_country_code}"
+        self.client.get(path)
+        for asset_path in get_asset_paths():
+            self.client.get(asset_path)
+
```
stevejalim, line 94: Note that I haven't put in a function to try out search on a Posts or Events listing page. 

We could add this, but the results so far implied it would just run up against the rate limiter if abused, else be OK

Issue #1694, opened 2020-07-10
Support additional hostname to smooth domain changeover


stevejalim:
AC
- [ ] updated release-branch code to allow both old and new domains under `settings.ALLOWED_HOSTS`

PR #1696, opened 2020-07-10
1694: Add support for responding to a third hostname, as part of prep…

stevejalim:
This is a temporary change, as part of prearations for a domain switchover, and will be backed out once the domain change is done.

APP_SUPPLEMENTARY_HOST is configured as a new domain for prod, and as an existing domain for stage and dev (because we don't have additional hostnames pointing at them). It's a required value, so we had to put something in there.

APP_SUPPLEMENTARY_URL is built from APP_SUPPLEMENTARY_HOST and is exposed as SUPPLEMENTARY_URL via env vars.

ALLOWED_HOSTS gets a SUPPLEMENTARY_URL value added to it. If that duplicates what BASE_URL or CDN_URL are set to, that does NOT break the site.

(Related issue #1694)


stevejalim:
@limed Cleanup of hostname handling ticked as a follow-up - see #1704

Issue #1709, opened 2020-07-13
Header logo and main menu


schalkneethling:
## Describe the bug

Some cleanup needed in the header with regards to the logo as well as the `Products & Technologies` as well as the `Communities` top-level menu entries.

## Possible solutions

- [ ] The logo has a background image set on the anchor element but, also an inline version using an image element. We can replace the inline image with something like `<span class="visually-hidden">Mozilla Developer</span>` and stick with the background image

- [ ] We also have the background logo image defined in two places as: `.header .mzp-c-navigation-logo a` and `.mzp-c-navigation-logo a`. One of these can be cleaned up.

- [ ] The top-level entries for both Products & Technologies and Communities have a class called `.mzp-c-menu-title` with a `::before` pseudo-element adding a base64 SVG version of the text already present.
    - Not sure if there is a valid reason for the above, but seems to me like we can safely remove the base64 image
- [ ] Add "hidden" accessibility nav to all pages

schalkneethling:
## Accessibility nav

### HTML

```
<ul id="nav-access" class="nav-access">
  <li>
    <a id="skip-main" href="#content">Skip to main content</a>
  </li>
  <li>
    <a id="skip-language" href="#language">Select language</a>
  </li>
  <li>
    <a id="skip-search" href="#main-q">Skip to search</a>
  </li>
</ul>
```

### SASS

```
/*
Keyboard & screen reader skip link menu
********************************************************************** */

.nav-access {
    width: 100%;
    position: absolute;
    top: -20em;
    z-index: 1001;

    a {
        position: absolute;
        left: 0;
        right: 0;
        background-color: rgba(255, 255, 255, .9);
        padding: 12px 10px;
        font-weight: bold;
        text-align: center;

        &:hover,
        &:focus {
            box-shadow: 3px 3px 5px #aaa;
            top: 20em;
            text-decoration: none;
        }
    }
}
```

stevejalim:
> Not sure if there is a valid reason for the above, but seems to me like we can safely remove the base64 image

If removed, we lose the close-trigger on expanded sub-menu in mobile viewport mode, so leaving it as it currently is. Hope that's cool.

PR #1717, opened 2020-07-14
1709: Header a11y improvements

stevejalim:
This changeset addresses a11y issues with the main nav spotted by @schalkneethling (thanks!)

* remove unnecessary 
* SCSS cleanup
* add quick-access nav links

Note that one recommendation from the source ticket #1709 was do drop a base-64 data-uri image from the CSS, but it looks like this was actually needed as the close trigger for the nav when in mobile-viewport mode.

Screenshot 2020-07-14 at 18 03 14

## How to test

- Code will be on [staging](https://developer-portal.stage.mdn.mozit.cloud/) by the time you read this

### DIFF
--- a/developerportal/templates/molecules/accessibility-nav.html
+++ b/developerportal/templates/molecules/accessibility-nav.html
@@ -2,9 +2,7 @@
   <li>
     <a id="skip-main" href="#content">Skip to main content</a>
   </li>
-  {% comment 'not rendered until search is released' %}
   <li>
     <a id="skip-search" href="#nav_search_query">Skip to search</a>
   </li>
-  {% endcomment %}
 </ul>
--- a/developerportal/templates/molecules/accessibility-nav.html
+++ b/developerportal/templates/molecules/accessibility-nav.html
@@ -1,0 +1,10 @@
+<ul id="nav-access" class="nav-access">
+  <li>
+    <a id="skip-main" href="#content">Skip to main content</a>
+  </li>
+  {% comment 'not rendered until search is released' %}
+  <li>
+    <a id="skip-search" href="#nav_search_query">Skip to search</a>
+  </li>
+  {% endcomment %}
+</ul>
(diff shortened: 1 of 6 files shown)


stevejalim:
@escattone I cross-browser tested, but Browserstack wasn't letting me get to localhost on iOS, so if you don't mind checking that, that'd be great

src/css/molecules/header.scss
```diff
@@ -1,25 +1,23 @@
-.header {
-  // FIX required for mzp
-  .mzp-c-navigation-menu {
-    box-sizing: unset;
-  }
+// FIX required for mzp
+.mzp-c-navigation-menu {
+  box-sizing: unset;
```
schalkneethling, line 3: Why did we need this override?

schalkneethling: I see. It affects the position of the main navigation in relation to the logo.

developerportal/templates/molecules/accessibility-nav.html
```diff
@@ -0,0 +1,10 @@
+<ul id="nav-access" class="nav-access">
+  <li>
+    <a id="skip-main" href="#content">Skip to main content</a>
+  </li>
+  {% comment 'not rendered until search is released' %}
```
stevejalim, line 5:
```suggestion
```

developerportal/templates/molecules/accessibility-nav.html
```diff
@@ -0,0 +1,10 @@
...
+  <li>
+    <a id="skip-search" href="#nav_search_query">Skip to search</a>
+  </li>
+  {% endcomment %}
```
stevejalim, line 9:
```suggestion
```

developerportal/templates/molecules/accessibility-nav.html
stevejalim: We can uncomment this section now we have search released

PR #1697, opened 2020-07-10
HPA Update for prod and celery-workers

limed:
Changeset revolves around HPA's in general, enables HPA for celery
workers and also tweaking max replicas for wagtail deployment for prod

(Resolves #1609)

## Key changes:

- Create HPA for `celery-worker`
- Update prod max replica for wagtail

Issue #1609, opened 2020-06-22
Prepare autoscaling (HPA) ahead of hard launch


stevejalim:
Would be great if hard launch means doing literally nothing in terms of dialling up resource etc :o)

Goal date for this is middle of July, before Friday 17th 

To do:

- [x] Estimate expected traffic @stevejalim / @valgrimm 
- [x] Set up Horizontal Pod Autoscaling @limed
- [x] Consider whether we need DB or cache scaling too (@stevejalim + @limed)
- [x] Explore using [AWS WAF](https://aws.amazon.com/waf/) around the CDN

stevejalim:
Estimated traffic wise is TBC, but load testing showed that with the CDN we can do 121RPS on a 2-4 pod Stage easily enough, so I suspect the current prod config is plenty

stevejalim:
@limed Do you think it's worth scaling up the DB, too? (So it can handle more connections -- tho rate limiting should stop abuse/floods)

PR #1728, opened 2020-07-15
Increase max pod for web pods

limed:
Increase prod web pods hpa and also rework Makefile targets, rework of PR #1697 

(Related issue #1609)

## Key changes:

- Increase max prod web pod  from 8->10
- Rework Makefile targets to allow more flexibility in the future

### DIFF
--- a/k8s/Makefile
+++ b/k8s/Makefile
@@ -193,7 +193,8 @@ k8s-jenkins-delete-rbac:
 .PHONY: k8s-ns k8s-delete-ns k8s-services k8s-delete-services \
 		k8s-deployments k8s-delete-deployments k8s-rollout-status \
 		k8s-wagtail-deployments k8s-celery-deployments \
-		k8s-deployment-hpa \
+		k8s-hpa k8s-deployment-hpa \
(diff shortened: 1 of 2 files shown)


k8s/Makefile
```diff
@@ -134,14 +134,18 @@ test-k8s-celery-beat-deployments:
...
 
-test-k8s-hpa:
+test-k8s-hpa: test-k8s-deployment-hpa
+
+test-k8s-deployment-hpa:
 	j2 hpa.yaml.j2 | kubeval --strict --ignore-missing-schemas
 
-k8s-delete-hpa:
-	${KC} delete --ignode-not-found deploy ${APP_NAME}
+k8s-delete-deployment-hpa:
```
stevejalim, line 147: Does this need to go in `PHONY`?

Plus where is/was `k8s-delete-hpa` in `PHONY`?

limed: Added in 40c39c198a25295c836be1e74e14a49bbb7b9e6f, there was never a `k8s-delete-hpa` target iirc

Issue #1719, opened 2020-07-15
Featured sections


schalkneethling:
The featured section reads like a list of featured items and should be marked up as such. Utilizing the existing classes and some of the structure we can amend the current HTML to something like:

```
<ul class="mzp-l-card-third">
  <li>
    <div class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9">
      ...
    </div>
  </li>
  <li>
    <div class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9">
      ...
    </div>
  </li>
  <li>
    <div class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9">
      ...
    </div>
  </li>
</ul>
```

valgrimm:
re-used in many places on site

stevejalim:
@schalkneethling Looking at the source Django template for this, it can be a list of rows of cards

eg 

```
[ 1, 2, 3 ]
[ 4, 5, 6] 
```

so am thinking of marking it up as a `ul` root with a list of sub-lists:

```
<ul class="featured-items-wrapper">

  {% split_featured_items featured as split_featured %}

  {% comment %}


  `split_featured` returns an iterable of dicts where each dict represents a list of
  items to render on a particular row
  [
      {
          items: <List>
          count: <int>
      },
      ...
  ]

  {% endcomment %}

  {% for row in split_featured %}
    <li>

    {% if row.count == 2 %}
    <ul class="mzp-l-card-half">
      {% for block in row.items %}
        <li class="mzp-c-card mzp-c-card-medium mzp-has-aspect-3-2">
          {% include "organisms/partials/featured-card-selector.html" with aspect_ratio="3_2" %}
        </li>
      {% endfor %}
    </ul>
    {% endif %}

    {% if row.count == 3 %}
      <ul class="mzp-l-card-third">
        {% for block in row.items %}
            <li class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9">
              {% include "organisms/partials/featured-card-selector.html" %}
            </li>
        {% endfor %}
      </ul>
    {% endif %}

    {% if row.count == 4 %}
      <ul class="mzp-l-card-quarter">
        {% for block in row.items %}
            <li class="mzp-c-card mzp-c-card-extra-small mzp-has-aspect-16-9">
              {% include "organisms/partials/featured-card-selector.html" %}
            </li>
        {% endfor %}
      </ul>
    {% endif %}

  </li>
  {% endfor %}
[...]

PR #1737, opened 2020-07-16
1719: Improve markup of Featured Items component for improved a11y

stevejalim:
This changeset removes all divs from the markup in favour of unordered lists. Now have a list of sublists, which represents both how the content is structured in terms of heirarchy and still flows appropriately.

(Related issue #1719)

## How to test

- Code is deployed to stage and you can see it on the Homepage and on any page in Products & Technologies

### DIFF
--- a/developerportal/templates/organisms/featured.html
+++ b/developerportal/templates/organisms/featured.html
@@ -19,7 +19,7 @@ Two items:
 
 {% endcomment %}
 
-<div class="featured-items-wrapper">
+<ul class="featured-items-wrapper">
 
   {% split_featured_items featured as split_featured %}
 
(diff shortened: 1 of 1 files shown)


schalkneethling:
On https://developer-portal.stage.mdn.mozit.cloud/topics/firefox-mobile/ I notice that the featured items is marked up as below:

```
<ul class="featured-items-wrapper">
    <li>
      <ul class="mzp-l-card-third">
        <li class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9"></li>
        <li class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9"></li>
        <li class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9"></li>
      </ul>
  </li>
</ul>
```

We have a list of 1 item that then contains another list of three items. I wonder, could we flatten it to just:

```
<ul class="featured-items-wrapper">
  <li class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9"></li>
  <li class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9"></li>
  <li class="mzp-c-card mzp-c-card-small mzp-has-aspect-16-9"></li>
</ul>
```

I edited the HTML in the browser and visually it would be fine, not sure if it is problematic from a template logic perspective though.

@stevejalim Thoughts?

stevejalim:
@schalkneethling Thanks for querying that - i get what you mean. The featured-items component can render 2 to 7 items depending on what's configured and splits them according to some opinionated logic.

eg: for five items, we will have

```
[   1   ]  [   2   ]
[ 3 ] [ 4 ] [ 5 ] 
```
which is rendered as a list of two rows, and each row is itself a list of items. I thought that seemed fairly logical/sensible in markup terms, but perhaps not.

What you're spotting here is where there is just one row of items to render, yet we still ahve that outer wrapping `ul` (ie a `li` for just one row, with its own `ul` and `li`s inside)

I think i could clean up the logic in [the template](https://github.com/mdn/developer-portal/blob/c65dd4e2627a4552b61fbc7f829f0c3704e1be27/developerportal/templates/organisms/featured.html) so that if there is only one row to render, we don't need to nest a `ul` inside a `ul`, though it will make the template logic a little busier.

Happy to try, particularly if screen reader/a11y support is hampered in this case.

Question: are you happy with the main list with sublists approach when there _are_ two rows of featured items, though? eg on  https://developer-portal.stage.mdn.mozit.cloud/topics/rust/ (which I've added more items to)

schalkneethling:
@stevejalim Makes sense, let's not overcomplicate this.

Issue #1720, opened 2020-07-15
Topics section


schalkneethling:
The topics list should be marked up as an unordered list. There is also an incorrect heading hierarchy.

A suggested restructuring could look something like:

```
<section id="topics" aria-labelledby="topics-main-heading">
  <h2 id="topics-main-heading">Browse products &amp; technologies</h2>

  <ul class=" mzp-l-card-quarter">
    <li class="mzp-c-card mzp-c-card-extra-small">
      <a class="mzp-c-card-block-link" href="/topics/browser-extensions/">
        <h3 class="mzp-c-card-title"><span>Add-Ons</span></h3>
        ...
      </a>
    </li>
    <li class="mzp-c-card mzp-c-card-extra-small">
      <a class="mzp-c-card-block-link" href="/topics/browser-extensions/">
        ...
      </a>
    </li>
    ...
  </ul>
</section>
```

Not the change in heading elements in the above code snippet.

stevejalim:
I think this will be automatically handled by #1719

PR #1747, opened 2020-07-25
1720: Topics component a11y improvements

stevejalim:
This changeset contains markup improvements to boost a11y, as suggested in #1720.

## How to test

- Code is being deployed to staging, so please check it out there. 
- There are three places the topics component is used:
  - homepage
  - /topics/
  - on a Person page, where they are associated with one or more topics/technologies. In this situation, the topics component renders its output as a two-column layout over four grid columns, not four cols over four.

### DIFF
--- a/src/css/molecules/item-topic-link.scss
+++ b/src/css/molecules/item-topic-link.scss
@@ -8,15 +8,15 @@ body.topics .topic-links .mzp-c-card {
 .topic-links {
   background-color: $color-light-gray-10;
 
-  h3 {
+  h2 {
     margin-top: $spacing-sm;
   }
-  h3,
-  h4 {
+  h2,
+  h3 {
     font-family: $body-font-family;
     font-size: 1.5rem;
   }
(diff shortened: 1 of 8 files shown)


src/css/molecules/item-topic-link.scss
```diff
@@ -8,15 +8,15 @@ body.topics .topic-links .mzp-c-card {
...
-  h4 {
+  h2,
+  h3 {
     font-family: $body-font-family;
     font-size: 1.5rem;
   }
-  h4 {
+  h3 {
     font-size: 1.5em;
   }
```
schalkneethling, line 21: Looks like this is already set on line 15 above?

schalkneethling:
@stevejalim On the `/topics/` page we do not have an `h1` element at all. Would it be possible to make the `topics-heading` on this page be a `h1`?

stevejalim:
@schalkneethling All fixed up

Issue #1758, opened 2020-07-28
Make `rel` attr of external links consistent


stevejalim:
Quick sweep to ensure that `nofollow` and `noreferer`are  in place where appropriate (in addition to existing `noopener`)

PR #1760, opened 2020-07-28
1758: Ensure all external links have the same values for the rel attribute

stevejalim:
For links that use `target="_blank"` we already had `rel="noopener"` set and sometimes `nofollow` and sometimes `noreferrer`, but this changeset brings them all into line so that ALL of them use `nofollow noreferrer noopener` for consistency and predictability. (I've also seen this pattern used by Mozilla Hacks blog.)

This does raise questions about SEO and analytics, regarding `nofollow` and `noreferer` and we could in the future support fine-grained control over those values for links.

Please feel free to disagree!

(Related issue #1758)

## How to test

- code will be on staging soon, so look at external links or social sharing links. However, what's more important is consensus on whether `nofollow` is definitely worth it and whether we are comfortable with/without `noreferrer`

### DIFF
--- a/developerportal/templates/organisms/newsletter-signup.html
+++ b/developerportal/templates/organisms/newsletter-signup.html
@@ -27,7 +27,7 @@
           <p class="newsletter-privacy-wrapper">
             <label for="newsletter-privacy" class="mzp-u-inline">
               <input type="checkbox" name="privacy" id="newsletter-privacy" required aria-required="true">
-              I'm okay with Mozilla handling my info as explained in this <a class="privacy-link" href="https://www.mozilla.org/privacy/websites/" target="_blank" rel="noopener noreferrer">Privacy Notice</a>.
+              I'm okay with Mozilla handling my info as explained in this <a class="privacy-link" href="https://www.mozilla.org/privacy/websites/" target="_blank" rel="noopener">Privacy Notice</a>.
             </label>
           </p>
           <p>
--- a/developerportal/apps/common/templates/button_block.html
+++ b/developerportal/apps/common/templates/button_block.html
@@ -6,7 +6,7 @@
       {% if value.page_link %}
         <a href="{% pageurl value.page_link %}" class="mzp-c-button">
       {% elif value.external_link %}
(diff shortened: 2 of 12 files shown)


developerportal/templates/organisms/newsletter-signup.html
```diff
@@ -27,7 +27,7 @@ <h2>Discover the best of web development</h2>
           <p class="newsletter-privacy-wrapper">
             <label for="newsletter-privacy" class="mzp-u-inline">
               <input type="checkbox" name="privacy" id="newsletter-privacy" required aria-required="true">
-              I'm okay with Mozilla handling my info as explained in this <a class="privacy-link" href="https://www.mozilla.org/privacy/websites/" target="_blank" rel="noopener noreferrer">Privacy Notice</a>.
+              I'm okay with Mozilla handling my info as explained in this <a class="privacy-link" href="https://www.mozilla.org/privacy/websites/" target="_blank" rel="noopener">Privacy Notice</a>.
```
schalkneethling, line 30: @stevejalim Is there a specific reason the link to the privacy policy is the odd one out?

schalkneethling: Ah! because this is one of the rare instances that we *do* know that we are linking to an external site we are comfortable endorsing.

stevejalim: My logic was: given we control both sides of the interaction (ie, link from a Mozilla property to another Mozilla property) there's no risk from including the referrer and it could be useful to analytics for the target side. 

However, I'm happy to change it! 😄

schalkneethling: Nope, I 100% agree. Took me a second to put two and two together 😄

stevejalim:
@valgrimm May I grab a view on this from you as PO?

valgrimm:
Need Chris' take on this on Thursday, deferring to his judgement as discussed

valgrimm:
Chris and I talked today about this and he said that it is fine to leave off the noreferrer header so other Moz properties know we are sending them traffic.

stevejalim:
For now, in the circumstances, am merging this as-is, and the rel attr can be loosened off in the future, if need be

PR #1765, opened 2020-07-29
1751: Extend app-level ratelimiting so that admin/CMS editors get a higher threshold

stevejalim:
This changeset updates the ratelimiting middleware to apply a less strict limit on requests made by authenticated staff users (ie, those with Wagtail CMS access). This is so that things like the image search typeahead isn't blocked so readily.

The initial value for the admin-user rate limiting is now 60 requests per minute, up from 25 per minute (which remains the default for unauthenticated users).

Note that the user has to be both authenticated AND a staff/admin user to not be treated as a regular public user. At the moment, all our authenticated users are staff, but this may not always be the case.

Tests have been extended to prove this is working as expected, plus manual checks have been done, but it's still possible that we need to increase the admin-user limit further, or simply remove it and keep it only for unauthenticated users.

## How to test

- code is deployed to staging
- as an unauthenticated user (eg in an incognitio window) repeatedly load a non-CMS page from staging - eg the homepage. It should rate limit on the 26th request, if you're fast enough
- as an authenticated user, in admin, you are now allowed more requests per minute, which should not get in the way of your editing/content work. Please check things that previously tripped the rate limiter. 
  - The image-search typeahead will, for instance, still hit it if you enter 61 characters one by one

@valgrimm Am still quite tempted to remove the rate limiting for authenticated staff users altogether -- I know you were keen to keep it and only relax it, but having implemented this to bucket users as either public or authenticated staff, we could drop the limit altogether for the staff users and we'd still be running rate limiting against someone unauthenticated pointlessly hammering at the admin login screen

### Snippets from manual testing:

Public rate limiting still working:
```
In: for i in range(30): print(i+1, requests.get("http://localhost:8000"))
1 <Response [200]>
2 <Response [200]>
3 <Response [200]>
4 <Response [200]>
5 <Response [200]>
6 <Response [200]>
7 <Response [200]>
8 <Response [200]>
9 <Response [200]>
10 <Response [200]>
11 <Response [200]>
12 <Response [200]>
13 <Response [200]>
14 <Response [200]>
15 <Response [200]>
16 <Response [200]>
17 <Response [200]>
18 <Response [200]>
19 <Response [200]>
20 <Response [200]>
21 <Response [200]>
22 <Response [200]>
23 <Response [200]>
24 <Response [200]>
25 <Response [200]>
26 <Response [429]>
27 <Response [429]>
28 <Response [429]>
29 <Response [429]>
30 <Response [429]>
```

Admin rate limiting does kick in, just after more queries:

Screenshot 2020-07-29 at 12 55 16

### DIFF
--- a/developerportal/apps/common/middleware.py
+++ b/developerportal/apps/common/middleware.py
@@ -19,7 +19,6 @@ def _get_group(request):
 def _get_appropriate_rate(group):
     if group == RATELIMIT_GROUP_ADMIN_REQUESTS:
         return settings.DEVPORTAL_RATELIMIT_ADMIN_USER_LIMIT
-
     return settings.DEVPORTAL_RATELIMIT_DEFAULT_LIMIT
 
 
--- a/developerportal/apps/common/middleware.py
+++ b/developerportal/apps/common/middleware.py
@@ -6,6 +6,22 @@ from ratelimit import ALL
 from ratelimit.core import is_ratelimited
 from ratelimit.exceptions import Ratelimited
 
+RATELIMIT_GROUP_PUBLIC_REQUESTS = "public_requests"
+RATELIMIT_GROUP_ADMIN_REQUESTS = "admin_requests"
+
+
+def _get_group(request):
+    if request.user and request.user.is_authenticated and request.user.is_staff:
+        return RATELIMIT_GROUP_ADMIN_REQUESTS
+    return RATELIMIT_GROUP_PUBLIC_REQUESTS
+
+
+def _get_appropriate_rate(group):
+    if group == RATELIMIT_GROUP_ADMIN_REQUESTS:
+        return settings.DEVPORTAL_RATELIMIT_ADMIN_USER_LIMIT
+
+    return settings.DEVPORTAL_RATELIMIT_DEFAULT_LIMIT
+
 
 def rate_limiter(get_response):
     """Enforces rate-limiting on all views.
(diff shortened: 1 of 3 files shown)


developerportal/apps/common/middleware.py
```diff
@@ -6,6 +6,22 @@
...
+def _get_group(request):
+    if request.user and request.user.is_authenticated and request.user.is_staff:
+        return RATELIMIT_GROUP_ADMIN_REQUESTS
+    return RATELIMIT_GROUP_PUBLIC_REQUESTS
+
+
+def _get_appropriate_rate(group):
+    if group == RATELIMIT_GROUP_ADMIN_REQUESTS:
+        return settings.DEVPORTAL_RATELIMIT_ADMIN_USER_LIMIT
+
```
stevejalim, line 22:
```suggestion
```

developerportal/apps/common/middleware.py
```diff
@@ -6,6 +6,21 @@
...
+RATELIMIT_GROUP_ADMIN_REQUESTS = "admin_requests"
+
+
+def _get_group(request):
+    if request.user and request.user.is_authenticated and request.user.is_staff:
+        return RATELIMIT_GROUP_ADMIN_REQUESTS
+    return RATELIMIT_GROUP_PUBLIC_REQUESTS
+
+
+def _get_appropriate_rate(group):
```
Gregoor, line 19: nit: wondering if a more appropriate (`UnintendedPunException`) would be: `_get_ratelimit_group` for the first one, and `_get_ratelimit_rate` for the second one?

